PR #6, opened 2017-07-10
Changes

psicotropicos:
- Removing react-native-share duplication
- Adding react-native-tilt


jimmylee:
:+1: LGTM, thank you thank you!

PR #8, opened 2017-07-11
Add some packages

GertjanReynaert:


jimmylee:
Whoa @GertjanReynaert thanks for adding all of these! Could you do me a small favor and double check if these work with Expo and/or the web? Thanks a bunch!!

GertjanReynaert:
Totally fair request! Added the necessary support booleans

jimmylee:
LGTM :) thank you so much @GertjanReynaert

PR #9, opened 2017-07-11
Add RNFirebase

designorant:


jimmylee:
@designorant hey thanks a ton for doing this, would it be possible for you to check if the library works with Expo and web, then updating the object data? Thanks.

designorant:
@jimmylee I don't use Expo or react-native-web but RNFirebase uses Firebases' native modules so that's a no for both Expo and the web.

However, there's also a lightweight [Firebase Web SDK](https://www.npmjs.com/package/firebase) written in pure JavaScript so maybe we could add that to the list as well?

[Differences explained](https://github.com/invertase/react-native-firebase#why).

jimmylee:
@designorant I think its totally fine that its a no for `expo` and `web`, I just want to reduce the amount of questions someone may ask given the context:

For example, I am an Expo dev, I can now know to avoid this library if I am attached. I also now know that if I detach to ExpoKit, I can use this library.

Thank you so much for revisiting your PR, I'll make sure to land this :)

PR #10, opened 2017-07-11
Add react-native-ble-plx

Cierpliwy:


jimmylee:
@Cierpliwy hey thanks for doing this! Small favor to ask: can you check if these libraries work with Expo and web and update the JSON object data? thanks :)

Cierpliwy:
Hello! This module is based on native modules for Android and iOS so obviously it won't work on web. Regarding expo I think that I shouldn't mark it as supported. Normal expo apps don't work well with custom native modules without detaching project to ExpoKit, which is basically a conversion to plain react native project. Also user must configure XCode/AndroidStudio project manually to properly setup permissions for Bluetooth.

jimmylee:
@Cierpliwy its totally fine that the libraries don't work with Expo :)

I just want an `expo: false` on the JSON object so everyone can know without having to ask :)

Cierpliwy:
Updated

jimmylee:
LGTM :) thank you!

watadarkstar:
@jimmylee It would be nice if you had a flag that indicated if the app worked with expo or required you to detach the expo app. For example, this library requires you to detach the app.

PR #14, opened 2017-07-12
Add tipsi-stripe

isnifer:


jimmylee:
@isnifer thank you so much for adding this! If you could update your PR to say `expo:false` that would be great. Since Tipsi-stripe only works when detached.

I'll make this more clear in the Readme, sorry about this.

isnifer:
@jimmylee oh, ok, no problem

jimmylee:
Thank you so much! Will land this soon :)

PR #21, opened 2017-07-17
Remove duplicated react-native-social-share

giautm:
Duplicated module from lines https://github.com/expo/react-native-libraries/blob/master/react-native-libraries.json#L247-L251


jimmylee:
@giautm thank you so much! LGTM.

PR #24, opened 2017-07-17
Remove duplicated react-native-tcp

giautm:
Duplicated from lines https://github.com/expo/react-native-libraries/blob/master/react-native-libraries.json#L211-L215


jimmylee:
:+1: LGTM, thanks a ton @giautm

PR #25, opened 2017-07-17
Remove duplicated react-native-cookies

giautm:
Duplicated from lines https://github.com/expo/react-native-libraries/blob/master/react-native-libraries.json#L242-L245

Sorry about noisy PR, this is the last duplicated. (Before we have CI for auto check)


jimmylee:
@giautm no worries :) Thank you !

PR #26, opened 2017-07-17
Added react-native-simple-dialogs and react-native-keyboard-manager

douglasjunior:
I added in the middle to prevent conflicts at the end of the file.


jimmylee:
I'm going to land this PR, and close your other one. Thanks for adding these libraries @douglasjunior  :)

PR #29, opened 2017-07-19
Added react-native-fs

itinance:
Adding react-native-fs


jimmylee:
LGTM :) thank you so much!

PR #31, opened 2017-07-30
Mark collapsible as maintained

oblador:
The project has a new maintainer and would say it's fairly active for it's size at the moment: https://github.com/oblador/react-native-collapsible/pulse/monthly


jimmylee:
:+1: LGTM, thanks for the update :)

PR #35, opened 2017-08-01
Remove duplicate of react-native-pdf-view

davidkhuu:
Removing duplicate entry of `react-native-pdf-view`

Appears on line [291](https://github.com/react-community/native-directory/blob/95afe29a916d21d3707efc3975cfabb0776478d5/react-native-libraries.json#L291) and [1181](https://github.com/react-community/native-directory/blob/95afe29a916d21d3707efc3975cfabb0776478d5/react-native-libraries.json#L1181) of the `react-native-libraries.json` file at the time of this commit. 

I removed the latest entry.

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -1178,11 +1178,6 @@
     "android": true
   },
   {
-    "githubUrl": "https://github.com/cnjon/react-native-pdf-view",
-    "ios": true,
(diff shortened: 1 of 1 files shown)


jimmylee:
:+1: LGTM, I don't know what I would do without you :)

Issue #20, opened 2017-07-15
pushedAt is not reliable


kelset:
### Description
In the `LibraryListItemColumnThree` component, to calculate the "Updated at" field, to the `getTimeSinceToday` method the value `props.library.github.stats.pushedAt` gets passed.

But this field (provided by the GitHub API) is actually not "reliable": what I mean by that is that the field `pushedAt` is not the datetime of the last commit pushed to master - which to me should define when the repo has been updated.

To prove it I'll use the [`react-native-push-notification`](https://github.com/zo0r/react-native-push-notification) repo - if you check via native.directory ([here](https://www.native.directory/updated/push-notifications)) you can see that it says that it has been updated for the last time "2 weeks ago" because it uses

>  "pushed_at": "2017-06-30T17:58:21Z",

as datetime to use for the calculation.

But if you check instead the [commit history](https://github.com/zo0r/react-native-push-notification/commits/master) it will say that the last commit to master was done on May 15th (`2017-05-15T18:43:43Z`) - which is way more than 2 weeks ago.

I can't actually figure out where the GitHub API gets that value for pushed_at, but yeah it's not the last commit to `master`.

### Proposed solution

While searching around, I managed to use the [GraphQL explorer](https://developer.github.com/v4/explorer/) from GitHub itself to search (in APIv4) for the last commit date using this query:

```js
query { 
  repository(owner:"zo0r", name:"react-native-push-notification") {  
    ref(qualifiedName: "master") {
      target {
        ... on Commit {
          id
          committedDate
        }
      }
    }
  }
}
``` 

So my proposal would be to use, instead of the APIv3 (REST) for GitHub, actually move to its v4 and, by using GraphQL, fetch all the necessary data plus the datetime of the last commit to master, and then use that one to calculate the "Updated XXXX ago" field.

But it may be "a big deal" so I wanted to open an issue to discuss with you guys if it would be a good idea or if it would be better to stick with v3 - in that case, I guess it would be necessary a secondary GET to the API for the last commit for each repo.

If you greenlight it I'd love to try to do it and PR (but I can't give any ETD atm).

jimmylee:
I'm a fan of GraphQL and so are the members on our team. So if you want to commit the time to creating a PR to migrating our code over to GraphQL I'm in.

Thanks for thinking about this! Really appreciate you helping us move forward.

kelset:
I think you should thank @brentvatne for pointing me to the repo ;)
[...]

PR #38, opened 2017-08-02
Implement a new scoring system

### LINKED ISSUE #33: Suggestion for project quality metrics
I think this website is a really well-needed resource and thanks for creating it.

One of the key bits for me is the project health feature. I think the simplistic score you have currently is a great start, but I was reminded of the great project quality feature that the Cocoapods (iOS dependency manager) directory has. You can see an example of a libraries quality page here:

https://cocoapods.org/pods/AnimationSugar/quality

By default, it starts a library with a score of 50 and then adds and subtracts from that to get the final score. Things like "being very popular" get an addition and others like "using a GPL licence" (which is arguably incompatible with the Apple App Store) result in a subtraction from the base score.

On the page I linked above, press on the "show metrics not applied" to see the full list of possible metrics. They have also documented how this system works and the theory behind it here:

https://guides.cocoapods.org/making/quality-indexes

Overall, I think a system like this would be a great way to measure the health of React Native libraries as well. Let me know your thoughts and I'm happy to help out if this is a direction you want to move in.

matt-oakes:
I have implemented a new scoring system based on the ideas from the [Cocoapods Quality Metrics](https://guides.cocoapods.org/making/quality-indexes). The new system takes into account:

* Popularity using the subscribers, forks, stars, and npm downloads with items at the start of the list weighted for greater importance as they indicate a more active contribution to the project.
* If the library is marked as recommended (called `goldstar` internally).
* If there are lots of open issues (currently set to 75).
* If there is no licence or a licence which is deemed incompatible with the Apple App Store.
* If the library has been recently updated (last 30 days) or not been updated in a while (over 6 months).

To help compare the previous and new scores, I have produced some statistics using [this script](https://gist.github.com/matt-oakes/55e8cb02d40555f083cb1a468923a96a):

```
Old mean:  73.82
New mean:  55.15
Mean difference:  -18.67

Number of libraries:  258
Match modifier counts:  {
  "Popular": 51,
  "Recently updated": 252,
  "No license": 60,
  "Lots of open issues": 14,
  "Very popular": 10,
  "Recommended": 10,
  "GPL license": 1
}
```

The average score is now lower, but I think this is correct. 50 should be the mark of an average quality library with lower and higher scores being indicative of the quality.

I have also put the data into a Google Sheet to allow you to check over the data more thoroughly:

https://docs.google.com/spreadsheets/d/1GIaIS_JIiz1ile5RknQPF6rjWFbfX8yRaYWB-vH-NQA/edit#gid=904868035

Looking through the scores, most seem to have been reduced because Github cannot determine which licence the library is under. I think this is a valid reason to reduce the score as this would be a barrier to using the library for most people.

All of these values and modifiers and be easily tweaked to fine tune the score.

Let me know what you think!

*Closes #33.*

### DIFF
--- a/scripts/build-and-score-data.js
+++ b/scripts/build-and-score-data.js
@@ -5,6 +5,7 @@ import jsonfile from 'jsonfile';
 import _ from 'lodash';
 import fetchGithubData from './fetch-github-data';
 import calculateScore from './calculate-score';
+import fetchLicense from "./fetch-license";
 import fetchReadmeImages from './fetch-readme-images';
 import fetchNpmData from './fetch-npm-data';
 
@@ -35,6 +36,10 @@ const buildAndScoreData = async () => {
   console.log('** Loading data from Github');
   await sleep(1000);
   let data = await loadRepositoryDataAsync();
+
+  console.log('** Fetching license type');
+  await sleep(1000);
+  data = await Promise.all(data.map(d => fetchLicense(d, d.github.fullName)));
 
   console.log('\n** Scraping images from README');
   await sleep(1000);
--- a/scripts/calculate-score.js
+++ b/scripts/calculate-score.js
@@ -1,63 +1,94 @@
+// This is an array of modifier objects. Each modifier has a name, value, and condition.
+// The condiction is passed the data and if it returns true, the value is added to the
+// libraries score.
+// 
+// The idea is to add to the score for good behaviour and subtract from the score for
+// bad behaviour. It is completely reasonable for no modifiers to match a given library
+// and it should get a score close to the middle. To help with this, the positive and
+// negative modifier scores should be roughly the same magnitude.
+const modifiers = [
+  {
+    name: "Very popular",
+    value: 40,
+    condition: data => popularityScore(data) > 10000
+  },
+  {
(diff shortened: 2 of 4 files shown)


jimmylee:
I'm so happy about this, thank you! I'm about to head to the airport and be out of pocket for about 11 hours. I am excited to review the work and land it into our codebase.

matt-oakes:
I have noticed that the issue from #20 is causing the *Recently updated* rule to be applied to libraries which shouldn't qualify. I have asked over in that issue if this is a task which @kelset is currently working on.

jimmylee:
TLDR: LGTM

One concern:

- I'm going to need to update the interface because users have mentioned in the past that having a score beneath 75% is a little upsetting. If I make it clear that 50% represents usability, then I think no one will feel bad.
- I need to make an information page about the scoring at one point.

Again thank you so much for doing this. Sorry it took so long for me to land it :) 🛩

matt-oakes:
@jimmylee No problem. That sounds right to me. It might just be changing it so it looks more like this:

```
|---------------------|---------------------|
Poor-----------------Good---------------Great
```

PR #47, opened 2017-09-01
Added web compatible libraries

peggyrayzis:
To help push adoption of universal components forward, I really want `native.directory` to be a one stop shop for people looking for RN libs that work on web.

I went ahead and added libs that I know for sure work on web (we used them in prod at MLS). Over the next few days, I'm going to be adding more once I write a script to determine a lib's web compatibility %.

I also added `react-native-create-bridge` to the directory. 😀


jimmylee:
@peggyrayzis you rock! Thank you so much

LGTM

PR #55, opened 2017-09-26
add react-native-zip-archive

plrthink:
react-native-zip-archive provides zip and archive utilities on both Android and iOS platform, it has tested on our product (https://mockingbot.com/) in production for a long time.


jimmylee:
Hey! Very cool :) Thanks for sharing

PR #68, opened 2017-10-22
Add a bunch of goodies

sonaye:
I use these in production.


jimmylee:
@sonaye thank you so much! I need to spend some time porting styles and then I will land this right after.

PR #74, opened 2017-11-01
add react-native-multiple-select lib

enieber:


jimmylee:
Nice another Expo library! Thank you so much @enieber

PR #77, opened 2017-11-06
added react-native-event-listeners

meinto:


jimmylee:
LGTM! Thank you so much @meinto!

PR #78, opened 2017-11-07
added react-native-viewpager-carousel & react-native-mirror

meinto:


jimmylee:
I like your libraries Tobias! Thanks for adding them 👍

PR #86, opened 2017-11-25
It doesn't work on Expo since it requires linking.

serranoarevalo:


jimmylee:
LGTM :) thank you so much!

PR #102, opened 2018-02-04
add face pile and async storage

peterpme:


jimmylee:
LGTM! Thanks a ton @peterpme

PR #114, opened 2018-03-12
react-native-google-play-game-services

Noitidart:
react-native-google-play-game-services. Android only for now. will update with ios if it gets added, there was hopeful talk of it


jimmylee:
@Noitidart thanks for submitting this :)

Noitidart:
Woohoo thanks!!

PR #116, opened 2018-03-15

gogoku:


jimmylee:
Hey @gogoku thanks for submitting, why doesn't this work with Expo?

gogoku:
@jimmylee sorry i had not used it in an expo built app, i just tried and it works with expo

jimmylee:
@gogoku sorry about the delay!

PR #119, opened 2018-03-21
Add Instabug for React Native to the directory.

azamzamy:
Instabug is an in-app bug reporting and automatic crash reporting SDK. It provides all device details, network logs, view hierarchy, and steps to reproduce with every bug or crash report.


jimmylee:
Awesome project! thank you so much @azamzamy !

PR #122, opened 2018-04-02
Autofocus the search input

fson:


jimmylee:
Hey @fson !! thanks for the PR

We're going to want to make sure this only happens on the web and not when you're viewing using a mobile browser (or else the keyboard appears and its annoying)

fson:
Hey, that's a great point. I tested on iPhone (with Safari) and on Android (with Chrome).

On mobile Safari focusing the field requires user interaction (to avoid the keyboard from appearing on many websites that do this), so this code has no effect whatsoever.

On Chrome for Android the field gets focus (a blinking cursor is shown), but the keyboard doesn't appear. That behavior is a bit funny, but good enough imho.

jimmylee:
Okie! Glad you tested this, I'm going to land this PR.

PR #123, opened 2018-04-04
Add react-native-keyguard

NishanthShankar:


jimmylee:
@NishanthShankar thanks for submitting this, have you tested this with Expo?

NishanthShankar:
This requires changes in the native code. So I don't believe it is expo
friendly

PR #124, opened 2018-04-27
add react-native-image-marker

JimmyDaddy:


jimmylee:
Cool library, thanks for sharing!

PR #132, opened 2018-06-10
Added react-native-super-ellipse-mask

everdrone:
source:

https://github.com/everdrone/react-native-super-ellipse-mask


jimmylee:
Looks cool @everdrone, thanks for the submission.

PR #141, opened 2018-09-11
Added React Native Snackbar

levani:
https://github.com/cooperka/react-native-snackbar


jimmylee:
Thanks Levani! You're a boss.

PR #143, opened 2018-09-16
Added react-native-date-picker

henninghall:
A date picker component for android and ios.


jimmylee:
@henninghall thank you so much for your contribution! :D

PR #145, opened 2018-09-17
Add react-native-gradients

yildizberkay:


brentvatne:
looks like this depends on only react-native-svg, so it works in expo

yildizberkay:
@brentvatne thank you very much for your suggestion, I updated my commit.

PR #146, opened 2018-09-18
add sideswipe, best carousel

JulianKingman:
Sideswipe is the best carousel for react native; lightweight and not plagued by bugs.


jimmylee:
Thank you so much @JulianKingman

PR #150, opened 2018-10-07
add react-native-gradient-buttons to list

thomaswangio:


jimmylee:
@thomaswangio thank you!

PR #157, opened 2018-12-09
add react-native-hotspot

lfkwtz:


jimmylee:
Thanks @lfkwtz for adding this!

PR #158, opened 2018-12-15
Added react-native-track-player

Guichaguri:
An audio player package dedicated to music/podcast apps.


jimmylee:
Nice one :) Thanks @Guichaguri

PR #159, opened 2018-12-21
Adding react-native-webview-readability onto list of packages

jeffreymendez1993:
A cross platform Safari like readerMode component that takes any url source that displays views cleanly with noise.


jimmylee:
Nice one @jeffreymendez1993 :-)

PR #160, opened 2018-12-21
Adding react-native-text-avatar onto list of packages

hicay:


jimmylee:
Nice addition :-) Thanks

PR #194, opened 2019-05-29
react-native-paper works on web too

ferrannp:
See https://callstack.github.io/react-native-paper/using-on-the-web.html


turnrye:
Thanks @ferrannp !

PR #228, opened 2020-02-13
Update build command

brentvatne:
Fetch updated libraries on each build

### DIFF
--- a/assets/data.json
+++ b/assets/data.json
@@ -1,6 +1,61 @@
 {
   "libraries": [
     {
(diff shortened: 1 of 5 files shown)


brentvatne:
this isn't great because we can easily get rate limited and as the number of libraries grows we'll be hammering github with requests in order to get star count etc. probably need to rethink the strategy of how data is being updated from github, maybe just do it on a schedule.

brentvatne:
This works pretty good, we'll probably want to actually commit library updates occasionally though. Good enough for now!

PR #229, opened 2020-02-14
Add Windows and macOS platforms

brentvatne:
# Summary

Microsoft is investing a lot in React Native for Windows and macOS. As their adoption grows outside of Microsoft it will be useful to see which libraries are supported on those platforms!

This PR adds filters but does not mark any libraries as compatible with either yet. Waiting on @stmoy to come up with a list of them before we merge this.

## Test Plan

Try the URL that is linked to below.


brentvatne:
lol dammit it seems i can't include `@` in my branch name or zeit has a hard time with it

PR #257, opened 2020-04-07
Add react-native-animated-spinkit

tienph6m:
# Why
Add react-native-animated-spinkit, a pure JavaScript port of SpinKit for React Native.

# Checklist

If you added a new library:

BTW, react-native-libraries.json doesn't seem to be formatted correctly. I have my editor with `Format on save` on, and I had to save without formatting so it wouldn't create a huge diff.

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -3364,5 +3364,18 @@
     "android": true,
     "expo": false,
     "npmPkg": "rn-zendesk"
+  },
+  {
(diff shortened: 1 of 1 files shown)


brentvatne:
ooh, can you submit a PR to fix the formatting @tienph6m?

tienph6m:
@brentvatne sure! Do you think we should also add a pre-commit hook like this  https://github.com/expo/expo-cli/blob/f9ef499532153b482f43acc87bee0e06fb336109/package.json#L22 ?

PR #261, opened 2020-04-09
Add pre-commit hook for linting and data validation

tienph6m:
# Why
As discussed here https://github.com/react-native-community/directory/pull/257#issuecomment-610545155, the `react-native-libraries.json` file is not being formatted correctly. 

I added a pre-commit hook with `lint-staged` to run eslint and prettier on only staged files.
Also, I noticed we have a schema file to validate `react-native-libraries.json`, so I added the command to validate it to `lint-staged` too.


brentvatne:
nice :) thanks a lot 🙏

PR #268, opened 2020-05-06
Mark react-native-smart-badge Library Unmaintained

bell-steven:
# Why

This PR marks the react-native-smart-badge library as "unmaintained."

The library hasn't been updated since November 2016. It is not compatible with any of the recent version of RN because it relies on PropTypes from React-Native.  There are 3 open PRs from 2017 that would fix the issue, but none have been merged.

# Checklist
N/A


brentvatne:
thanks!

PR #269, opened 2020-05-06
Added Parse SDK package

espipj:
# Why

Added Parse SDK for the [Parse Platform](https://parseplatform.org/)
- I _think_ it might work on Expo, and Web as the SDK was initially focused on vanilla JS (for web) but I haven't used it on those. Can confirm that it **works on iOS and Android**.

# Checklist

If you added a new library:

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -3679,5 +3679,15 @@
     "windows": false,
     "unmaintained": false,
     "npmPkg": "react-native-voip-push-notification"
+  },
+  {
+    "githubUrl": "https://github.com/parse-community/Parse-SDK-JS",
+    "ios": true,
+    "android": true,
+    "web": false,
+    "expo": false,
+    "windows": false,
+    "npmPkg": "parse",
+    "unmaintained": false
   }
 ]
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -3684,8 +3684,8 @@
     "githubUrl": "https://github.com/parse-community/Parse-SDK-JS",
     "ios": true,
     "android": true,
-    "web": false,
-    "expo": false,
+    "web": true,
+    "expo": true,
     "windows": false,
     "npmPkg": "parse",
     "unmaintained": false


brentvatne:
hey there! I'm not sure about adding this one, it's not really specific to react-native. maybe we need to do some more thinking about what packages belong here

brentvatne:
also - if it works on ios and android without needing you to install separate native modules then it works in expo client also

react-native-libraries.json
```diff
@@ -3679,5 +3679,15 @@
     "windows": false,
     "unmaintained": false,
     "npmPkg": "react-native-voip-push-notification"
+  },
+  {
+    "githubUrl": "https://github.com/parse-community/Parse-SDK-JS",
+    "ios": true,
+    "android": true,
+    "web": false,
```
brentvatne, line 3687: if this was made originally for web should it not work on web?

brentvatne, line 3687:
```suggestion
    "web": true,
```

react-native-libraries.json
```diff
@@ -3679,5 +3679,15 @@
...
+    "expo": false,
```
brentvatne, line 3688:
```suggestion
    "expo": true,
```

brentvatne:
I think we need to be selective about the libraries that we add that aren't *specifically for React Native*, and include them only if 1) they are popular 2) it will help clarify any ambiguity around whether the library is compatible

espipj:
Hey @brentvatne ! Actually you are right, I am not 100% sure if this should be included, as it is not specific package for react-native like `react-native-firebase`, altough it has an specific interface [parse/react-native](https://github.com/parse-community/Parse-SDK-JS/blob/master/react-native.js)

espipj:
Oh, seems that I am late 😅

espipj:
> also - if it works on ios and android without needing you to install separate native modules then it works in expo client also

Yes it does, I didn't write down Expo compatible as I've never tried it with Expo myself

Issue #271, opened 2020-05-07
Add thumbnail feature


jonsamp:
# Feature Request

## Why it is needed

It's great to see a thumbnail of images from the repo inline. The directory used to have this feature, but it was removed when refactoring/redesigning. [This tweet](https://twitter.com/Gorhom/status/1258134417387724806?s=20) also requested it.

## Possible implementation

- Add icons that are clickable/hoverable that display a thumbnail from the repo.
- There's a `fetchReadmeImages` in **build-and-score-data.js** that can be used to fetch images again.

## Design

Here's a mock up of what this could look like:

<img width="902" alt="Screen Shot 2020-05-06 at 8 47 57 PM" src="https://user-images.githubusercontent.com/6455018/81242499-4f580600-8fdb-11ea-8c01-516ac2d63e94.png">

And here's an svg for the icon:

```
<svg width="16" height="16" fill="none" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M14 2H2v12h12V2zM0 0v16h16V0H0z" fill="#82889E"/><path fill-rule="evenodd" clip-rule="evenodd" d="M7 7.586l3 3 2-2 3.707 3.707-1.414 1.414L12 11.414l-.586.586 2.293 2.293-1.414 1.414L7 10.414l-5.293 5.293-1.414-1.414L7 7.586z" fill="#82889E"/><path d="M12 5.5a1.5 1.5 0 11-3 0 1.5 1.5 0 013 0z" fill="#82889E"/></svg>
```

gorhom:
hi @jonsamp, thanks for your fast respond and opening up an issue for it.

i was thinking of adding a new field in the json schema `thumbnails` < an array of image urls, the reason why not scraping images from the readme, because that not all the time readme will include a previews of the library itself, there might be a tutorial , logo and sponsors images.

let me know what do you think ?

jonsamp:
I think that's a great idea, and would really make the thumbnails valuable! When we had them before many of them were sponsor logos etc. Sounds like this would fix that!

PR #276, opened 2020-05-07
feat: add thumbnails

gorhom:
# Why

closing #271

# Preview

On Desktop

ezgif-4-cba2f6e6d938

On Mobile

ezgif-4-aea6f8cd5cb2

# Test

you will need to add a library with thumbnails field 

```json
[
  ...,
  {
    "npmPkg": "@gorhom/animated-tabbar",
    "githubUrl": "https://github.com/gorhom/react-native-animated-tabbar",
    "ios": true,
    "android": true,
    "web": false,
    "expo": true,
    "examples": [
      "https://github.com/gorhom/react-native-animated-tabbar/tree/master/example"
    ],
    "images": [
      "https://raw.githubusercontent.com/gorhom/react-native-animated-tabbar/master/docs/previews/bubble.gif",
      "https://raw.githubusercontent.com/gorhom/react-native-animated-tabbar/master/docs/previews/flashy.gif"
    ]
  },
  ...,
]
```

# Checklist


brentvatne:
looks great @gorhom! we should be able to turn [this script](https://github.com/react-native-community/directory/blob/master/scripts/fetch-readme-images.js) back on and populate it with images from the readme if we want to do this in an automated way. what do you think?

gorhom:
@brentvatne yeah it will be very helpful, but my concern was that not all images in readme are preview of the library.

However, we could let libraries' owner to set `thumbnails`, and as a fallback we run the `fetchReadmeImages` script ?

gorhom:
> However, we could let libraries' owner to set thumbnails, and as a fallback we run the fetchReadmeImages script ?

@brentvatne 🙂

brentvatne:
oops I thought I had responded @gorhom - sorry!

I think we can let people provide `thumbnails` and if they do we don't scrape them from README. Otherwise we fallback to README scraped images.

gorhom:
> oops I thought I had responded @gorhom - sorry!
no worries,

>I think we can let people provide thumbnails and if they do we don't scrape them from README. Otherwise we fallback to README scraped images.
edit: maybe we can use a field called images instead? it would be more clear for folks who don't speak english natively

sounds good, im on it👍

gorhom:
@brentvatne it's ready 👍

PR #286, opened 2020-05-07
Added 22 new packages

wbroek:
# Why

Added 22 new packages to the library ; not my own (just one) but use them on daily basis

# Checklist
If you added a new library:

- [X ] Added it to **react-native-libraries.json**

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -516,13 +516,6 @@
     "expo": false
   },
   {
-    "githubUrl": "https://github.com/zo0r/react-native-push-notification",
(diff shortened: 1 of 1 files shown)


brentvatne:
hello! something seems off with this diff - it seems like some libraries are being replaced. can you ensure that nothing is being removed and only new libraries are added? thanks

Screen Shot 2020-05-07 at 3 38 22 PM

wbroek:
Thanks @brentvatne for pointing that out; my json was older then the one from master. Fixed that now and removed the duplicates as well

PR #300, opened 2020-05-09
Add topics and make searching by topics simpler

FarazPatankar:
I have displayed the topics associated with each library and also made them clickable. This way, whenever a user clicks a topic, they can search by that specific topic.

I also created a small demo to see the feature in action.
add-topics

### DIFF
--- a/components/Library/index.tsx
+++ b/components/Library/index.tsx
@@ -5,6 +5,7 @@ import { isEmptyOrNull } from '../../util/strings';
 import { colors, layout, A, Label, Caption } from '../../common/styleguide';
 import { Badge } from '../Icons';
 import { CompatibilityTags } from '../CompatibilityTags';
+import { Topics } from '../Topics';
 import { MetaData } from './MetaData';
 
 type Props = {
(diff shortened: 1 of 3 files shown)


brentvatne:
hello! thanks for the pr! after seeing this in the ui i'm not totally sure we want it to be featured so prominently. i'll defer to @jonsamp for design ideas here.

FarazPatankar:
Haha, fair enough. I thought the page could use some colors but I am open to changing whatever.

I was also considering using icons for the platforms but I didn't know what to use for MacOS. 🤔

brentvatne:
Another thing to consider for these tags - currently we're just using "topics" from GitHub, but this doesn't work too well in some situations, eg: if a library is inside of a monorepo, it can't have its own topics. So we should maybe use npm keywords in addition to GitHub topics and `uniq` them.

FarazPatankar:
I'd be open to that. Are we already fetching that information?

brentvatne:
not yet - we're just fetching download stats from npm: https://github.com/react-native-community/directory/blob/master/scripts/fetch-npm-data.js

jonsamp:
This is a cool feature! Thanks for adding it. Since there are many topics and they act like site links, I think we can style these like links. Here's a design that puts topics below the description. The text-size here is 12px (and the background and border is the same as the link styling):

Screen Shot 2020-05-15 at 10 42 05 AM
Screen Shot 2020-05-15 at 10 44 34 AM

FarazPatankar:
Ooh, I like that. I'll try to update and push this over the weekend!

brentvatne:
lovely @jonsamp!

PR #318, opened 2020-05-29

asklar:
Add some Windows-supported modules

# Why

This adds a few modules that are supported on ReactNative Windows

# Checklist

If you added a new library:

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -374,7 +374,6 @@
     "android": true,
     "web": false,
     "expo": true,
-    "windows": true
   },
   {
     "githubUrl": "https://github.com/prscX/react-native-shine-button",
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -373,7 +373,7 @@
     "ios": true,
     "android": true,
     "web": false,
-    "expo": true,
+    "expo": true
   },
   {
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -373,7 +373,8 @@
     "ios": true,
     "android": true,
     "web": false,
-    "expo": true
+    "expo": true,
+    "windows": true
```
Simek, line 377: Hi @asklar! I think that you have in mind `@react-native-community/async-storage` package, not this one, right?

asklar: yep good catch, I'll change it :)

react-native-libraries.json
```diff
@@ -3280,9 +3281,15 @@
...
     "expo": true,
-    "windows": false,
+    "windows": true,
     "unmaintained": false
   },
+  {
+    "githubUrl": "https://github.com/microsoft/react-native-dualscreen",
+    "android": true,
+    "windows": true,
+    "npmPkg": "https://www.npmjs.com/package/react-native-dualscreeninfo"
```
Simek, line 3291: Oh, I have just spotted that `npmPkg` here should be just the package name (`react-native-dualscreeninfo`).

PR #320, opened 2020-05-30

asklar:
# Why
turns out we have some more to in order to fully support dualscreen on Windows
(sorry for the noise :) )

CC @kmelmon


brentvatne:
np, thanks for updating!

PR #322, opened 2020-05-30
sort platforms alphabetically

Simek:
# Why

This PR sorts the platforms in the alphabetical order in the filters and libraries boxes. The currently used order was a bit ambiguous and may seems a bit random (form the code I can assume that it was just based on the platforms addition time). The alphabetical sort is fair and intuitive for the users, the same approach has been introduced on the RN website not that long ago.

I was thinking about updating the Readme file and JSON schema too but I think that the order in there is not that crucial. Of course feel free to correct me if I'm wrong.

# Checklist

If you added a new library:

# Preview
Annotation 2020-05-30 134400

### DIFF
--- a/components/CompatibilityTags.tsx
+++ b/components/CompatibilityTags.tsx
@@ -11,12 +11,12 @@ type Props = {
 export function CompatibilityTags(props: Props) {
   const { library } = props;
   const platforms = [
-    library.ios ? 'iOS' : null,
     library.android ? 'Android' : null,
-    library.windows ? 'Windows' : null,
+    library.expo && typeof library.expo !== 'string' ? 'Expo client' : null,
+    library.ios ? 'iOS' : null,
     library.macos ? 'macOS' : null,
     library.web ? 'Web' : null,
-    library.expo && typeof library.expo !== 'string' ? 'Expo client' : null,
+    library.windows ? 'Windows' : null,
   ]
     .map(platform => platform)
     .filter(Boolean);
--- a/components/Filters.tsx
+++ b/components/Filters.tsx
@@ -19,12 +19,20 @@ type FilterButtonProps = {
 
 const platforms = [
   {
+    param: 'android',
+    title: 'Android',
+  },
+  {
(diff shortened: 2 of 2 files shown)


brentvatne:
now that I look at the original sorting I think that I basically just sorted in order in which the platform support for react-native was released, and moved Expo client to the end because it felt materially different from the other options.

I like that with the current sorting we group together [ mobile ] [ web ] [ desktop ] [ other ], I think this is a bit more useful thank alphabetical sorting. what do you think?

Simek:
I have noticed a group after a while, but at the first glimpse it was not that obvious, but in the other hand those groups make a sense and I think that Android and iOS should be at the beginning because those platforms are build-in.

I think that it will be nice to communicate or mark those groups somehow, I have prepared a mockup, don't look at icons I have use the random ones, but the most important thing there is the concept:

directoryplatforms

I don't know if adding the icons is the best solution, maybe some other visual grouping will be better to make ii more obvious for the users.

I'm also not that big fan of check marks as a visual addition, when they have no functional aspect. 
It would make more sense to me when all platform will be always visible (first two rows below box). 
The last two rows shows the grouping concept and labels without check marks (which look a bit cleaner to me).

Simek:
@jonsamp What's your opinion on that?

jonsamp:
@Simek @brentvatne I'm a fan of just alphabetical sorting with no icons, because i feel like the lines between mobile/tablet/desktop/web are blurry (iPads are laptops and MS Surfaces are tablets). So my vote is just to sort them like @Simek did originally.

With that, if you both feel strongly about icons (or separation on platform screen size), we'd want to put the icons in a horizontal or vertical row/column. If you look at the right section of the existing site, having the icons aligned makes it easier to discern visually:

Screen Shot 2020-06-09 at 12 25 01 PM

I can make a mock up if you want. I think we'd probably want to copy the existing details section picture above, then have two columns of information on each library result (one column with platform compatibility, and one with the library's meta data)

PR #50, opened 2017-09-10
[WIP] Query GrahpQL GitHub Endpoint

kelset:
### DO NOT MERGE YET

I've decided to start reporting this PR earlier because it has many moving parts and I'd prefer to make sure I'm going in the right direction before going too far - it started with the discussion in #20 (that will get closed after this PR will get merged I guess).

#### Recap of things that happens

* a quick prettier + typos fix for all files in `./scripts`
* Added a new implementation of the `fetch-github-data.js` logic that uses  `apollo-fetch` to query the GraphQL GitHub endpoint
* Added a new field `lastMasterCommitData` that is basically a more reliable timestamp to evaluate when a lib has been updated (check issue for more details)
* To query the new GraphQL API an OAuth token is necessary, so I had to add to my local `secret.json` a new field `GITHUB_OATH_TOKEN` that I have generated via my profile.

#### Issues

* a couple fields, `hasPages` and `hasDownloads`, are not available via the GraphQL endpoint (or, at least, I couldn't find them) - but it also looks like they are not used (AFAIK)
* when a lib is not "available" or something (haven't investigated properly), the endpoint returns `null`, reason why I had to add a check in the `fetchGithubDataViaGraphQL` method and filter after `loadRepositoryDataAsync` in `build-and-score-data.js`
* There is no easy way to parse all topics per repo (AFAIK), need to set an upper bound number (now it's 10)

#### TODO

- [ ] fix topics parsing
- [ ] move fetch-license to graphql (or even in the same call as fetch-github-data)
- [ ] use the new `lastMasterCommitData` in the score calculation
- [ ] ??? (*suggestions*)

PR #333, opened 2020-06-21
Fix updated at time to use default branch commits and check monorepo directories

### LINKED ISSUE #20: pushedAt is not reliable

alexbrazier:
fixes #20
# Why

The updated time was including any updates to the repo, such as issues created or commits in other branches. I think users would expect the updated time to either represent the last time a commit was made to the default branch e.g. master or the last publish time.

The updates would always be the same for monorepos, so every time large ones like expo had any updates the first few pages would be expo projects, even though only one of them had been updated.

# Changes

- Uses graphql api which allows us to make a single request to fetch the package.json and other details
- Fixes updated time so it only checks commits on the default branch
- Get the correct time for monorepos as it checks for commits within the folder

# Checklist

Screenshot showing expo monorepo now has different updated dates that matches the folder location and master branch
image

Thanks to @kelset for providing initial graphql query in #50


now.json
```diff
@@ -1,8 +1,7 @@
 {
   "build": {
     "env": {
-      "GITHUB_CLIENT_ID": "@github_client_id",
-      "GITHUB_CLIENT_SECRET": "@github_client_secret"
+      "GITHUB_TOKEN": "@github_token"
```
alexbrazier, line 4: I don't think it's possible to use a client id and secret with the graphql api so this needs to be updated to a [github token](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line).

scripts/fetch-github-data.js
```diff
@@ -49,99 +100,101 @@ export const fetchGithubRateLimit = async () => {
   };
 };
 
-// const repoRequestCache = {};
-
-export const fetchGithubData = async data => {
+const getUpdatedUrl = async url => {
```
alexbrazier, line 103: Sometimes the graphql API doesn't pick up redirects. This function gets the updated url using the rest API to retry.

scripts/fetch-github-data.js
```diff
@@ -49,99 +100,101 @@ export const fetchGithubRateLimit = async () => {
...
-      createdAt: json.created_at,
-      pushedAt: json.pushed_at,
-      forks: json.forks,
-      issues: json.open_issues,
-      subscribers: json.subscribers_count,
-      stars: json.stargazers_count,
+      hasIssues: json.hasIssuesEnabled,
+      hasWiki: json.hasWikiEnabled,
+      hasPages: json.deployments.totalCount > 0,
+      hasDownloads: true,
```
alexbrazier, line 184: I couldn't find this in graphql, and I'm not really sure what it is meant to represent. I've set it to `true` to maintain backwards compatibility as every repository listed in the data was `true`

types/index.ts
```diff
@@ -45,9 +45,9 @@ export type Library = {
     license: {
       key: string;
       name: string;
-      spdx_id: string;
+      spdxId: string;
```
alexbrazier, line 48: These two fields have been updated by graphql. I couldn't see them being used anywhere but it made more sense to just leave them in camelcase to match the rest of the output rather than having to modify them. I can revert this if it's going to cause a breaking change.

scripts/fetch-github-data.js
kelset: I think it probably refers to the repo having files/installers attached to releases, but not sure tbh.

kelset:
hey Alex, thank you so much for picking this back up 👍

@brentvatne I think this new approach is totally worth as the original problem highlighted in my issue seems to still stand.

brentvatne:
i pulled the branch locally and i had trouble fetching data using `yarn data:update` with the current batch size of 100 - i believe the graphql api has different throttling rules. i reduced the batch size to 50 and that resolved the issue for me.

we should update `fetchGithubRateLimit` to use the graphql api to determine [how many more node requests are available](https://developer.github.com/v4/guides/resource-limitations/) and also compare that with the number of nodes required per repo request * number of repos to bail out of updating data and deploying if we are throttled

brentvatne:
merged in https://github.com/react-native-directory/website/commit/4c1b864e616fdb5e13813acffb908b3611cc1727 - thanks a lot! if you can follow up with a fix for `fetchGithubRateLimit` and so on as described in previous comment that'd be amazing

PR #334, opened 2020-06-21
add "Recently Added" libraries sort order

### LINKED ISSUE #91: Order by Recent Added
Sometimes out of curiosity I want to check which is the recent one added in the list. It is not same as Last Updated.

Simek:
# Why

Fixes #91

This PR adds "Recently Added" sort order to the website. To achieve this I had to remove the sorting while updating and processing data and regenerate `data.json` file. 

The other way to make this possible was to add an additional field to the processed entity like `initialOrder` which should be populated based on the unmodified `react-native-libraries.json` but the solution above seems for me as more clear.

The label and the position in the Sort component can be changed, it were just my first ideas.

> P.S. I'm going to fix all the lint errors in the separate PR later, I just want to make the changes of this PR clear (so I have omitted the commit hook).

### Preview
Untitled

# Checklist

If you added a new library:


brentvatne:
looks like this needs to be rebased after https://github.com/react-native-directory/website/commit/4c1b864e616fdb5e13813acffb908b3611cc1727 :)

Simek:
@brentvatne Done!

PR #338, opened 2020-06-23
Use graphql api rate limit and calculate points per request

### LINKED ISSUE #337: Fix GitHub rate limit checks for GraphQL API
In https://github.com/react-native-directory/website/pull/333 we switched over to using the GitHub GraphQL API, which has a different mechanism for rate limiting. We should update our data update scripts to account for this.

alexbrazier:
closes #337

# Why

Uses the correct graphql rate limit endpoint to match the graphql query we make to fetch github info.

It makes a dummy request using actual query as github will return the actual `cost` of the query rather than having to manually calculate it and update it when the query changes.

The query actually only turns out to cost `1` point, and we get `5000` an hour (same as the REST api)

Output
```
  yarn data:update                            
yarn run v1.21.1
$ babel-node scripts/build-and-score-data.js --presets @babel/preset-env
** Loading data from GitHub
4993 of 5000 GitHub API requests remaining for the hour at a cost of 1 per request
```

### DIFF
--- a/scripts/build-and-score-data.js
+++ b/scripts/build-and-score-data.js
@@ -146,7 +146,7 @@ async function loadRepositoryDataAsync() {
     githubResultsFileExists = true;
   } catch (e) {}
 
-  let { apiLimit, apiLimitRemaining } = await fetchGithubRateLimit();
+  let { apiLimit, apiLimitRemaining, apiLimitCost } = await fetchGithubRateLimit();
 
   // 5000 requests per hour is the authenticated API request rate limi
   if (apiLimit < 5000) {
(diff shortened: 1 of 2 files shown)


alexbrazier:
@brentvatne I've now made the fix to the rate limit check.

It turns out the query is only worth 1 point, and I verified this by checking the requests only went down by 1 at a time, so I'm not sure why you were having an issue with requesting 100 at a time.

brentvatne:
thanks for this!

yeah i'm not quite sure what happened with the batch size either, we can play with this more or wait for it to become too slow as library count grows and deal with it then.

PR #340, opened 2020-06-24
display package license information if provided

Simek:
# Why

This PR adds licence row to the `MetaData` column. Since the license information is included in the directory score calculation it would be nice to display it. It's also a common information displayed for the GitHub repositories and NPM packages.

I have created the simple icon myself, it can be clanged in any way you like. For example GitHub uses scale weight icon for license. Also I'm not sure about the best placement so I have put this row at the end for now.

### Preview
Annotation 2020-06-24 192006

# Checklist

If you added a new library:

### DIFF
--- a/components/Icons/index.tsx
+++ b/components/Icons/index.tsx
@@ -224,6 +224,17 @@ export function Thumbnail({ width = 16, height = 16, fill = colors.gray3 }: Prop
   );
 }
 
+export function License({ width = 16, height = 18, fill = colors.black }: Props) {
+  return (
+    <Svg width={width} height={height} viewBox="0 0 16 18" fill="none">
+      <Path
+        fill={fill}
+        d="M0,0v18h16V0H0z M14,15.9H2V3.1h12C14,3.1,14,15.9,14,15.9z M12,13H4v-1h8V13z M12,10.5H4v-1h8V10.5z M12,8H4V7h8V8z"
+      />
+    </Svg>
+  );
+}
+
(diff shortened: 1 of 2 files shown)


brentvatne:
a couple comments:
- it looks like maybe licenses aren't fetched properly in monorepos - `expo-barcode-scanner` should be MIT licensed
- perhaps in the case of "Other" license we should say "Unrecognized License" or something like that

Simek:
> * it looks like maybe licenses aren't fetched properly in monorepos - `expo-barcode-scanner` should be MIT licensed
Yup, it looks like all monorepo packages are affected after the GraphQL refactor. I'm working on the fix for that.

> perhaps in the case of "Other" license we should say "Unrecognized License" or something like that
I like this idea, I'm going to adjust this text in the next commit, thank you!

alexbrazier:
> > * it looks like maybe licenses aren't fetched properly in monorepos - `expo-barcode-scanner` should be MIT licensed
> 
> Yup, it looks like all monorepo packages are affected after the GraphQL refactor. I'm working on the fix for that.
> 
> > perhaps in the case of "Other" license we should say "Unrecognized License" or something like that
> 
> I like this idea, I'm going to adjust this text in the next commit, thank you!

@Simek Just fixed it here https://github.com/react-native-directory/website/pull/342

jonsamp:
This is looking good. Nice addition! A UI tweak: can we move the license icon + value about the website icon + value?

Simek:
@jonsamp I have updated the rows order, thank you for the feedback!

PR #342, opened 2020-06-24
fix: Monorepo license issue by parsing package.json

alexbrazier:
# Why

Monorepos had a bug meaning licenses weren't used from the package.json. They also would have had the wrong format.

This change parses the package.json to get the license for monorepos and then replaces it with the github license scheme to match the others.

image

# Checklist

### DIFF
--- a/debug-github-repos.json
+++ b/debug-github-repos.json
@@ -4,6 +4,10 @@
     "expo": true
   },
   { "githubUrl": "https://github.com/gre/gl-react", "expo": true },
+  {
+    "githubUrl": "https://github.com/expo/expo/tree/master/packages/expo-barcode-scanner",
+    "expo": true
+  },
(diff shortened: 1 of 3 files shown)


alexbrazier:
Should I commit the updated `data.json`?

brentvatne:
@alexbrazier - it's not necessary, but in this case it could be helpful for local testing. generally this is just committed occasionally to have somewhat up to date data in dev environment. i sent you an invite to be a collaborator, feel free to merge when you're ready :)

Simek:
> Should I commit the updated `data.json`?

Usually not, in case of my "Recently Added" PR the data regenerate was necessary to make this feature work on DEV.

alexbrazier:
Thanks, I'll leave the `data.json` out of this one then as the structure hasn't changed

PR #347, opened 2020-06-25
add "Maintained" filter

Simek:
# Why

This PR introduces another group of filters - Status - and the first filter in this group which allows to hide libraries with `unmaintained` value set to `true`. I think that this new feature will be useful and besides that I have few other ideas related to the new filters in this group.

I'm not sure about naming, in the initial version I have used the `Hide unmaintained` label (because it was more obvious that when checkbox is not selected the results are not affected).

### Preview
Annotation 2020-06-25 152612

# Checklist

### DIFF
--- a/components/Filters.tsx
+++ b/components/Filters.tsx
@@ -113,6 +113,12 @@ export const Filters = (props: FiltersProps) => {
           ))}
         </View>
       </View>
+      <View style={styles.container}>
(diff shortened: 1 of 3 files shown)


Simek:
> I imagine "recommended" would be another good one here.

I agree, I have been also thinking about "Updated in" + select with "last week", "last month", "last year" etc.

PR #385, opened 2020-07-17
Redeploy app on a schedule

### LINKED ISSUE #355: Refresh library stats and info more frequently
Currently it seems the stats only get updated when the site gets redeployed. This means the site gets outdated quickly as the recently updated section is no longer accurate.

Is it possible to run the build script every so often on the server, e.g. every hour to match the github token?

alexbrazier:
Closes #355 

# Why

We display the time since updated on the website, but this only gets updated when the site gets re-deployed meaning it becomes inaccurate fairly quickly

This will trigger a redeploy on a schedule to update the site even if no commits have been merged to it

### DIFF
--- a/.github/workflows/deploy-to-production.yml
+++ b/.github/workflows/deploy-to-production.yml
@@ -3,6 +3,9 @@ name: Deploy to production
 on:
   push:
     branches: [ master ]
+  schedule:
+    # Run every 3 hours
+    - cron: '0 */3 * * *'
 
 jobs:
   build:
--- a/.github/workflows/deploy-to-production.yml
+++ b/.github/workflows/deploy-to-production.yml
@@ -4,8 +4,8 @@ on:
   push:
     branches: [ master ]
   schedule:
-    # Run every 3 hours
-    - cron: '0 */3 * * *'
+    # Run every 1 hour - https://crontab.guru/#0_*/1_*_*_*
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-to-production.yml
```diff
@@ -3,6 +3,9 @@ name: Deploy to production
 on:
   push:
     branches: [ master ]
+  schedule:
+    # Run every 3 hours
+    - cron: '0 */3 * * *'
```
alexbrazier, line 8: I'm wondering if we should even reduce this to run every hour? That way the update time will almost be completely accurate. It would also sync well with the github token rate limit reset time.

brentvatne: every hour makes sense!

brentvatne, line 8:
```suggestion
    - cron: '0 */1 * * *'
```

.github/workflows/deploy-to-production.yml
```diff
@@ -3,6 +3,9 @@ name: Deploy to production
 on:
   push:
     branches: [ master ]
+  schedule:
+    # Run every 3 hours
```
brentvatne, line 7:
```suggestion
    # Run every 1 hour
```

PR #388, opened 2020-07-18
Force new deploy even with no changes

alexbrazier:
- Triggers rebuild of package info

Follow up to #385 and #355

As there are no code changes the deployment gets skipped on the scheduled runs. This forces it to redeploy anyway.

```
now [options] <command | path>

  Options:

    -h, --help                     Output usage information
    -v, --version                  Output the version number
    ...
    -f, --force                    Force a new deployment even if nothing has changed
```


brentvatne:
nice, thanks :)

PR #394, opened 2020-07-20
add license fetch fallback based on package.json

Simek:
# Why

This PR adds license fallback for non-monorepo packages based on `package.json`. Fallback is only executed when there is no license fetched from GitHub or GitHub classified license as "other". This should fix most of issues related to the licenses.

# Preview
Annotation 2020-07-20 195033

Annotation 2020-07-20 195133

# Checklist

### DIFF
--- a/scripts/fetch-github-data.js
+++ b/scripts/fetch-github-data.js
@@ -247,7 +247,7 @@ const createRepoDataWithResponse = (json, monorepo) => {
     }
 
     if (!json.licenseInfo || (json.licenseInfo && json.licenseInfo.key === 'other')) {
-      json.licenseInfo = getLicenseFromPackageJson(packageJson);
+      json.licenseInfo = getLicenseFromPackageJson(packageJson) || json.licenseInfo;
     }
   }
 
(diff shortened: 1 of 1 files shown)


scripts/fetch-github-data.js
```diff
@@ -245,6 +245,10 @@ const createRepoDataWithResponse = (json, monorepo) => {
     if (json.topics.length === 0) {
       json.topics = packageJson.keywords;
     }
+
+    if (!json.licenseInfo || (json.licenseInfo && json.licenseInfo.key === 'other')) {
+      json.licenseInfo = getLicenseFromPackageJson(packageJson);
```
alexbrazier, line 250: Would you need to do a check here to make sure the license doesn't get set to `undefined` e.g. in the case where `key = 'other'` and the license is missing from the package.json?

Something like:
```suggestion
      json.licenseInfo = getLicenseFromPackageJson(packageJson) || json.licenseInfo;
```

Simek: Good point, thank you for the suggestion! It probably won't happen often, but it's good to have a fallback on `package.json` read result.

PR #405, opened 2020-07-24
add react native style cheatsheet in json file

luism3861:
# Why

this PR add react-native-styling-cheat-sheet in json file, this is not a dependency but this is very helpful to find all the css list you can use in React Native.!

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -6141,5 +6141,16 @@
     "windows": false,
     "macos": false,
     "unmaintained": false
+  },
+  {
+    "githubUrl": "https://github.com/vhpoet/react-native-styling-cheat-sheet",
+    "npmPkg": "react-native-styling-cheat-sheet",
+    "ios": true,
+    "android": true,
+    "web": true,
+    "expo": true,
+    "windows": true,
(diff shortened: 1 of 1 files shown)


Simek:
Hi @luism3861, thank you for preparing this PR!

I would like to see the development resources, tools or helpers in the directory, but for now it could be problematic to add a "library" that do not exist/is not published in to NPM registry and do not have a `package.json` file (directory data scraping, score calculation and display components depends on this data). 

I'm going to test this addition locally and check if there are any issues, but besides that, I will ask @brentvatne to review this PR too and present his point of view.

brentvatne:
this is great to have as a concrete real-world example as we design support for different types of resources within the directory. i believe we should hold off on landing it to master, but perhaps pull it into a development branch where @Simek has been experimenting with this type of thing

luism3861:
> this is great to have as a concrete real-world example as we design support for different types of resources within the directory. i believe we should hold off on landing it to master, but perhaps pull it into a development branch where @Simek has been experimenting with this type of thing

sounds great i think for different types of resources its a good experiment.!

PR #415, opened 2020-07-29
Updated react-native-sensitive-info entry

mCodex:
Included windows and macos compatibility in react-native-sensitive-info entry

If you added a new library:

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -5126,7 +5126,6 @@
     "web": false,
     "expo": false,
     "windows": true,
-    "macos": true,
     "npmPkg": "react-native-sensitive-info",
     "unmaintained": false
   },
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -5125,7 +5125,8 @@
     "android": true,
     "web": false,
     "expo": false,
-    "windows": false,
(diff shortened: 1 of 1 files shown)


mCodex:
@Simek thanks for pointing this out. I've already updated and released a new version including this fix.

Simek:
@mCodex No problem, but did you tested the library in the macOS project? Description is not mentioning macOS support too:
> react-native-sensitive-info manages all data stored in Android Shared Preferences, iOS Keychain and Windows Credentials. You can set and get all key/value using simple methods.

Currently lack of `TARGET_OS_OSX` in the iOS implementation suggest that there is no difference between iOS Keychain and macOS Keychain API, is it correct assumption? 

BTW. By adding `tvos` to the `podspec` you have enabled the use of library with tvOS platform. Did you test it in the `react-native-tvos` project?

Simek:
@mCodex I'm sorry, but your library code do not compile in the macOS project. I'm sad that I have to verify what library author says about the supported platforms. 

Screenshot 2020-07-29 at 21 41 10

mCodex:
Thanks for your help again! I'll close this for now and investigate that problem.

PR #453, opened 2020-09-14
add web support for react-native-msal

josmithua:
# Why

We've recently added web support to our package, although it's still in beta.
I also added the NPM package name


Simek (approved):
@josmithua Thank you for the update! 👍 Specifying `npmPkg` is not required when GitHub repository have the same name as npm package.

PR #483, opened 2020-10-15
added libraries

mtroskot:
# Why
Added few libraries

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -1,32 +1,5 @@
 [
   {
-    "githubUrl": "https://github.com/jhen0409/react-native-debugger",
-    "ios": true,
-    "android": true,
-    "expo": true
-  },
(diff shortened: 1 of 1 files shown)


Simek (changes requested):
Hi @mtroskot, few things:
* libraries should be added at the end of file
* `false` values can be omitted
* `npmPkg` can be omitted when package name is the same as GH repository
* `react-native-debugger` should be marked as `dev`

I'm also not sure if we can add `realm` at this moment, because we do not support minimal package version or scoped versions for them. According to the [Realm docs](https://docs.mongodb.com/realm/react-native/install/#install-realm-with-npm), you need to use `@beta` with React Native.

Simek (approved):
Thank you so much for addressing the review comments! LGTM 👍

PR #485, opened 2020-10-17
added tvOS

saideepesh000:
# Why

Added tvOS filter availability on the website. Click on tvOS filter if you want to filter out tvOS repos.

# Checklist

If you added a new library:

### DIFF
--- a/components/CompatibilityTags.tsx
+++ b/components/CompatibilityTags.tsx
@@ -46,6 +46,7 @@ export function CompatibilityTags(props: Props) {
     library.macos ? 'macOS' : null,
     library.web ? 'Web' : null,
     library.windows ? 'Windows' : null,
+    library.tvos ? 'tvOS' : null,
   ]
     .map(platform => platform)
     .filter(Boolean);
--- a/components/Filters.tsx
+++ b/components/Filters.tsx
@@ -45,6 +45,10 @@ const platforms = [
     param: 'windows',
     title: 'Windows',
   },
+  {
+    param: 'tvos',
+    title: 'tvOS',
+  },
 ];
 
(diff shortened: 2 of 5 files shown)


Simek (changes requested):
Hi @saideepesh000, thank you for working on this! 

Generally main obstacle for this feature is how detect and mark the existing libraries which supports tvOS (reading `.podspec` platform might be one of ways, but this might yields false-positives because libraries templates in the old days always were adding `tvOS` in there). There was a few other ideas and some discussions about this, but we have not get to any solid conclusion.

But adding this filter might be a first, good step. I have only one review request - please retain the alphabetic order of platforms. 🙂

saideepesh000:
> 
> 
> Hi @saideepesh000, thank you for working on this!
> 
> Generally main obstacle for this feature is how detect and mark the existing libraries which supports tvOS (reading `.podspec` platform might be one of ways, but this might yields false-positives because libraries templates in the old days always were adding `tvOS` in there regardless of actual support). There was a few other ideas and some discussions about this, but we have not get to any solid conclusion.
> 
> But adding this filter might be a first, good step. I have only one review request - please retain the alphabetic order of platforms. 🙂

@Simek, hey thanks for atleast accepting this PR with a review(I was bit nervous), this is my first time contributing to the community that I use(React Native).

Simek:
@saideepesh000 No need to be nervous 🙂 BTW. There is still an order issue - "Web", "tvOS", "Windows".

saideepesh000:
> 
> 
> @saideepesh000 No need to be nervous 🙂 BTW. There is still an order issue - "Web", "tvOS", "Windows".

😂, sorry for that

PR #504, opened 2020-11-15
Add the Formorama library

dezemand:
# Why
This library can be used in React Native to manage your forms. 

# Checklist
If you added a new library:


Simek (changes requested):
Hi @dezemand, I'm not able to see any mention about `react-native-web`, `react-native-macos` or `react-native-windows` in the docs or package repository. There is also no example, so I'm cannot accept the out-of-tree platforms support without the confirmation that they are indeed supported.

The example link should also lead to the working code example, not the code snippets in the docs.

Simek:
Closing due to lack of response. Please re-create the PR if you still want to add this library.

PR #525, opened 2021-01-05
Add windows support to react-native-permissions

jaimecbernardo:
# Why

Indicate Windows support in `react-native-permissions`.

The pre-commit hook also did some reformatting of **react-native-libraries.json**.

# Checklist

If you added a new library:


Simek (approved):
Thank you for the update! 👍

PR #526, opened 2021-01-06
Add Windows support to react-native-pdf

jaimecbernardo:
# Why

Indicate Windows support in `react-native-pdf`.

The pre-commit hook also did some reformatting of **react-native-libraries.json**.

# Checklist

If you added a new library:


Simek (approved):
Thank you! 🙂

PR #536, opened 2021-02-02
update Header component to redirect icon and text homepage

luism3861:
# Why
this PR update to redirect main page, icon and text in header component!., like react native docs you can redirect home page click both icon and text i think it's better!.


Simek (approved):
Hi @luism3861, sorry for the late review! 

It's good idea, I have tested the change locally and everything LGTM, thank you for this improvement! 👍

PR #546, opened 2021-02-18
attempt to speed up deploy but optimizing NPM retries

Simek:
# Why

This PR aims to speedup deployment process which is at this moment choked by the NPM retries. Currently all the NPM retry request were executed almost at the same time, which could irritate the NPM API. The idea behind the change is to assign the random timeout from given range and increase the `sleep` range with each try.

After the changes I was able to finish `yarn data:update` in several minutes, it time to try this on Vercel.

# Checklist

### DIFF
--- a/scripts/build-and-score-data.js
+++ b/scripts/build-and-score-data.js
@@ -24,8 +24,10 @@ const JSON_OPTIONS = {
   spaces: 2,
 };
 
-export const sleep = (ms = 0) => {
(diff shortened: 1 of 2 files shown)


Simek:
@brentvatne Unfortunately, the change won't help. Locally I have reach max 8 retries, while the data fetch on the Vercel reached 60 tries (and there were no visible time gains). Going to investigate this further.

PR #556, opened 2021-02-22
Add react-native-performance

oblador:
# Why

This PR adds react-native-performance library to the list. 

# Checklist

If you added a new library:

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -5861,5 +5861,15 @@
     ],
     "ios": true,
     "android": true
+  },
+  {
+    "githubUrl": "https://github.com/oblador/react-native-performance/tree/master/packages/react-native-performance",
+    "examples": ["https://github.com/oblador/react-native-performance/tree/master/examples/vanilla"],
+    "images": [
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -5861,5 +5861,15 @@
...
+  },
+  {
+    "githubUrl": "https://github.com/oblador/react-native-performance/tree/master/packages/react-native-performance",
+    "examples": ["https://github.com/oblador/react-native-performance/tree/master/examples/vanilla"],
+    "images": [
+      "https://user-images.githubusercontent.com/378279/105892056-9f677480-6011-11eb-895a-f3f8653449c8.png"
+    ],
+    "ios": true,
+    "android": true,
+    "expo": true
```
oblador, line 5873: I'm a bit unsure about this one. The library has a native component that is not available on expo, but it works without it.

Simek: It looks like your example converted to Expo Snack works fine (https://snack.expo.io/@simek/github.com-oblador-react-native-performance:examples-vanilla), so I see no reason not to mark Expo as supported. 

We can always revert it, if there will be some issues reported from Expo users. 🙂

Simek (approved):
@oblador LGTM, thank you for the addition! 👍

PR #578, opened 2021-03-01
Add @gabrielmoncea/react-native-template

gabrielmoncea:
# Why

Add a new library

# Checklist

If you added a new library:


Simek (approved):
Hi @gabrielmoncea, thank you for the addition. 👍 

I have removed the `dev` flag, because in the future we are planning to introduce the special tag/flag for the templates to differentiate them more from the libraries. 

P.S. It's only the second template entry, which will be added to the directory. 😉

PR #583, opened 2021-03-05
Added consistencss

mateosilguero:
# Why

Add [consistencss](https://github.com/mateosilguero/consistencss) library.

# Checklist

If you added a new library:

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -6082,11 +6082,7 @@
      "images": [
        "https://raw.githubusercontent.com/mateosilguero/consistencss/master/website/static/img/logo.png"
      ],
-     "ios": true,
      "android": true,
-     "web": true,
-     "expo": true,
-     "windows": true,
-     "macos": true
+     "ios": true
   }
 ]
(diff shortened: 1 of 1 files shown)


Simek (changes requested):
Hi @mateosilguero, due to compatibility issues in the past, we no longer accept out-of-tree platforms without a working demo or example (even if, from the source code, they are looking like they would work).

Also I was try to import the listed example into Snack, but it failed. 

If you want to see the library accepted quickly (without spending time on writing the examples) unfortunately you need to removed the out-of-tree platforms. Of course you will be able to add platform tags later in time.

react-native-libraries.json
```diff
@@ -6073,5 +6073,20 @@
...
+     ],
+     "images": [
+       "https://raw.githubusercontent.com/mateosilguero/consistencss/master/website/static/img/logo.png"
+     ],
+     "ios": true,
+     "android": true,
+     "web": true,
+     "expo": true,
+     "windows": true,
+     "macos": true
```
mateosilguero, line 6090:
```suggestion
     "android": true,
     "ios": true
```

Simek (approved):
Thank you for the review update! 👍

mateosilguero:
Hi @Simek, I understand, i removed the other platforms.

PR #589, opened 2021-03-07
Add @dietime/react-native-date-picker library

DieTime:
# Why

This library has a unified wheel interface for date selection, which makes it possible to have the same design on different platforms.

# Checklist

If you added a new library:


Simek (approved):
Hi @DieTime, thank you for the addition to the Directory! 👍

PR #596, opened 2021-03-12
add test deploy workflow, update setup-node

Simek:
# Why

It would be nice to have an ability to deploy test version of selected PRs. 

This changeset adds that ability, and will activate on PRs with the `deploy` label.

The action code is based on production script, but skips the `--prod` which should be enough, according to Vercel docs:
* https://vercel.com/docs/platform/deployments#vercel-cli

# Checklist

### DIFF
--- a/.github/workflows/test-deploy.yml
+++ b/.github/workflows/test-deploy.yml
@@ -1,0 +1,25 @@
+name: Test deploy
+
+on:
+  pull_request:
+    types: [ labeled ]
+
+jobs:
+  build:
+    if: ${{ github.event.label.name == 'deploy' }}
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-deploy.yml
```diff
@@ -0,0 +1,25 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js 14
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - name: Deploy
+        shell: bash
+        run: npx vercel --force --token "$VERCEL_TOKEN"
```
brentvatne, line 20: maybe we want to take the output from this and add it as a comment? not a big deal though as we can always click through to get the url

brentvatne (approved):

PR #598, opened 2021-03-13
add stora to react-native-libraries.json

chengsokdara:
# Why

This PR add `stora` in react-native-libraries.json file.

# Checklist

If you added a new library:

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -6259,15 +6259,19 @@
   },
   {
     "githubUrl": "https://github.com/dev-yakuza/react-native-image-modal",
-    "examples": [
-      "https://github.com/dev-yakuza/react-native-image-modal/tree/develop/Example"
-    ],
+    "examples": ["https://github.com/dev-yakuza/react-native-image-modal/tree/develop/Example"],
     "images": [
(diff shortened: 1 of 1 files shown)


Simek (changes requested):
Hi @chengsokdara, unfortunately it looks like the library you want to add doesn't work with Expo:
* https://snack.expo.io/@simek/amused-churros

I'm also not able to find a `web` platform example anywhere, for some time we do not accept out-of-tree platforms support without a confirmation that library supports them, even with library is not using the native components. This is due to compatibility issues reported by directory users.

If you want to proceed with addition you need to provide more examples or you can just remove platforms in question for now, and add additional support later in time.

chengsokdara:
@Simek Thank you for your response, I have solved conflicts and remove both web and expo from supported platform.

Simek (changes requested):
Please run `yarn libraries:format` script to drop the formatting changes (running Git Hooks on commit also should fix the problem), I don't have a permission to push the changes to your branch.

BTW. How you have ended with this reformatting in the changeset? I would like to fix this issue, if you are able to provide steps to reproduce.

chengsokdara:
@Simek Sorry it was me forgetting to yarn install so when I commit it didn't run husky. I did run yarn libraries:format just in case.

Simek (approved):
Thank you for the review updates! 👍

PR #599, opened 2021-03-13
Add macOS compatibility to react-native-tcp-socket

Rapsssito:
# Why

Adds macOS compatibility to react-native-tcp-socket

# Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -6165,7 +6165,8 @@
       "https://github.com/Rapsssito/react-native-tcp-socket/tree/master/examples/tcpsockets"
     ],
     "ios": true,
-    "android": true
+    "android": true,
+    "macos": true
   },
   {
     "githubUrl": "https://github.com/deanhet/react-native-text-ticker",


Simek (changes requested):
Hi @Rapsssito, we usually do not accept out-of-tree platforms without an example or confirmation, due to compatibility issues in the past.

In your library repository I can only see changes to the docs with mention of macOS support, but I did not see any specific macOS code and same applies for the `example` located in the repository. Also `react-native-macos` is not listed in any of dependencies groups, so at this moment I'm not able to accept the update.

Rapsssito:
@Simek, you are right. I just updated the docs, but if you take a look at the podspec and the changelog, the macOS compatibility came way before this change.

However, I understand the issue with the lack of examples. I will update the example folder as soon as possible and come back to you.

Simek (approved):
@Rapsssito Thank you for pointing out the code in Podspec, via blame, I was able to find the PR adding macOS support and take a look at this code. I think that is enough to accept the update. 🙂

PR #624, opened 2021-04-08
Speed up initial render experiment

Simek:
# Why

Currently the initial Directory page render could take a while, this is an attempt to speed it up using `next` [dynamic imports](https://nextjs.org/docs/advanced-features/dynamic-import) and `react-content-loader` SVG content skeletons.

# How

ATM the optimizations are applied only to the main page of Directory, the results are looking promising on the `localhost`, but the effect/benefit might differ in PROD (SSR) environment. Let's test it! 😄 

The solution also supports dark mode and should not degrade the functionality by any means.

# Preview

ooo

### DIFF
--- a/components/Explore/ExploreSection.tsx
+++ b/components/Explore/ExploreSection.tsx
@@ -1,3 +1,4 @@
+import dynamic from 'next/dynamic';
 import React, { useContext } from 'react';
 import { StyleSheet, View } from 'react-native';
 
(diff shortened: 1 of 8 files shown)


brentvatne (approved):
let's try it! worst case we just roll it back :)

Simek:
@jonsamp Feel free to adjust or correct the skeleton appearance later, for now I'm going to merge it as-is 🙂

PR #632, opened 2021-04-23
chore: add re-state

raulpesilva:
# Why

# Checklist

If you added a new library:

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -6502,16 +6502,10 @@
   },
   {
     "githubUrl": "https://github.com/raulpesilva/re-state",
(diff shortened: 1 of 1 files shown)


Simek (approved):
Hi @raulpesilva, thank you for adding your library! 👍 

Unfortunately due to issues in the past we do not accept out-of-tree platforms compatibility without an example project for that platform, however, I was able to verify that the package is compatible with Expo and RN Web using Snack:
* https://snack.expo.io/@simek/re-state

I have also removed the example link (Readme file), because it do not met the criteria:
> URLs to example projects (Snacks preferred) or with demonstrations of the library.

PR #635, opened 2021-04-25
update Readme

Simek:
# Why

The update to the Readme includes:
* RND logo
* section reorder
* fields description section reformat (fields linkable and visible in new GH ToC)
* `yarn` as preferred tool for Directory related operations
* minor text tweaks

# Preview

* https://github.com/react-native-community/directory/tree/readme-update

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,10 @@ This command creates site data in `./assets/data.json`
 GITHUB_TOKEN=<*> yarn data:update
 ```
 
+## How do I deploy to production?
+
+Get a commit on `master` and it will be automatically deployed.
+
 ## I don't like your website, can I hit an API instead and build my own better stuff?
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -129,10 +144,6 @@ GITHUB_TOKEN=<*> npm run data:update
 
 ```sh
 # once environment variables are configured, install vercel and deploy
-npm i -g vercel
+yarn global add vercel
```
brentvatne, line 147: these days i default to just explaining npm, because if someone is using yarn they know what the yarn equivalent is, but that may not be true the other way around (and explaining both seems like too much text).

that said, we probably want people to use yarn for this repo because of our lockfile, so maybe it makes sense to just say yarn here as well.

brentvatne (approved):

PR #660, opened 2021-06-07
Add react-native-big-list

marcocesarato:
# Why

This is a high performance list view for React Native with support for complex layouts using a similar FlatList usage to make easy the replacement. This list implementation for big list rendering on React Native works with a recycler focused on performance and memory usage and so it permits processing thousands items on the list.

# Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -6686,5 +6686,22 @@
     "ios": true,
     "android": true,
     "expo": true
+  },
+  {
+    "githubUrl": "https://github.com/marcocesarato/react-native-big-list",
+    "examples": [
+      "https://github.com/marcocesarato/react-native-big-list/tree/master/example", 
+      "https://marcocesarato.github.io/react-native-big-list/"
+    ],
+    "images": [
+      "https://raw.githubusercontent.com/marcocesarato/react-native-big-list/master/assets/screenshots/performance.gif",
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -6686,5 +6686,22 @@
...
+    "images": [
+      "https://raw.githubusercontent.com/marcocesarato/react-native-big-list/master/assets/screenshots/performance.gif",
+      "https://raw.githubusercontent.com/marcocesarato/react-native-big-list/master/assets/screenshots/example-section-list.jpg"
+    ],
+    "ios": true,
+    "android": true,
+    "web": true,
+    "expo": true,
+    "windows": true,
+    "macos": true
```
Simek, line 6705: Unfortunately we do not accept out-of-tree platforms without an example for those platforms, due to compatibility issues reported by the directory users earlier. You can update the supported platforms later, when you will add those platforms to the example.

Simek (commented):
Hi @maggialejandro thank you for adding new library to the directory! 👍 

Please address my comment below about out-of-tree platforms, other than that the entry LGTM! 🙂

react-native-libraries.json
marcocesarato: Ok understandable, I removed macos and windows from the package

Simek (approved):

PR #682, opened 2021-07-17
init concept of the search "Relevance" sort option

Simek:
# Why & How

Lately we have received the feedback about the libraries order, when user have entered the search keyword. So far, we just retain the order no matter what libraries API query has been sent. However, this might be a bit confusing, because users might suspect that after using a `search` query the results "are refined" to promote the best matched results (via weight or some kind of score).

To address that, and as an effort to improved the UX this PR introduces the concept of "Relevance" - the dynamic sort order,  which only appears in the sort option, if user have entered a query. 

I have also implemented a basic, automatic switch, so when users enter a serach term, the sort order switches automatically to "Relevance" (if user have not set earlier the sort order manually). there is still a small issue, if the user manually select the "Relevance" sort and then removed the search query, data is display correctly, but "Relevance" retains as select item value.

Also the `matchScore` calculation at this point is very basic (+100 for name match, +10 for description match, +1 for keyword/tag match), but I would like to retain low level of logic complexity in there, because this sorting is dynamical and cannot be easy cached, like the other sorts are. All the "Relevance" entries are sorted first by `matchScore`, then, if score matches by the default sort, which is "Recently Updated".

I have also included the debug data (`matchScore` value) at the top of the library box, it's just a temporary addition, to make the testing of the feature easier.

Please provide the feedback and your opinion abut this new feature and the way it has been implemented, and please treat this PR as PoC, this feature might be rewritten or not even finally land in the production. 😉 

# Test deploy

- https://react-native-directory-7jall9sn0-rndir.vercel.app/

# Checklist

CC: @satya164

### DIFF
--- a/components/Library/index.tsx
+++ b/components/Library/index.tsx
@@ -61,6 +61,9 @@ export default function Library(props: Props) {
           </View>
         ) : null}
         {showPopularity && library.popularity ? <PopularityMark library={library} /> : null}
+        {library.matchScore ? (
+          <Label style={{ color: colors.primary }}>Debug MatchScore: {library.matchScore}</Label>
+        ) : null}
         <View style={isSmallScreen ? styles.containerColumn : styles.displayHorizontal}>
           <A
             href={library.githubUrl || github.urls.repo}
--- a/components/Library/index.tsx
+++ b/components/Library/index.tsx
@@ -61,9 +61,6 @@ export default function Library(props: Props) {
           </View>
         ) : null}
         {showPopularity && library.popularity ? <PopularityMark library={library} /> : null}
-        {library.matchScore ? (
-          <Label style={{ color: colors.primary }}>Debug MatchScore: {library.matchScore}</Label>
-        ) : null}
         <View style={isSmallScreen ? styles.containerColumn : styles.displayHorizontal}>
           <A
             href={library.githubUrl || github.urls.repo}
--- a/components/Sort.tsx
+++ b/components/Sort.tsx
@@ -15,6 +15,10 @@ type SortButtonProps = {
 
 const sorts = [
   {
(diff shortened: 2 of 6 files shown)


Simek:
Test deploy URL:
* https://react-native-directory-kt122xz4h-rndir.vercel.app/

satya164 (approved):
Looks great 🙏

jonsamp (approved):
I like the logic behind this. Keeping this simple is smart.

Assuming you remove the debug match score, and we fix the sort selection issue where "relevance" appears and disappears, I think we should ship this. Just so others can see, here's video of that issue:


https://user-images.githubusercontent.com/6455018/126182702-985f54ba-1237-4011-bfcf-276c47dd5d88.mov

Thanks for making this PR!

Simek:
> Assuming you remove the debug match score, and we fix the sort selection issue where "relevance" appears and disappears, I think we should ship this. Just so others can see, here's video of that issue:

@jonsamp Thank you for the feedback! 🙂 So we should make the option always available in the `select`, but make it disabled when there is no search query, right?

brentvatne:
this is great!

what do you think about this solution to the `select` conundrum:
- make "Relevance" the default value in the `select`
- make "Relevance" behave the same as "Last Updated" when there is an empty string

this way we don't need to change the value of the `select` depending on the input. in a sense this is also accurate and expresses our intent - we believe that when there is no explicit search string then the most relevant results may likely be the most recently updated libraries.

Simek:
Update preview deploy URL:
* https://react-native-directory-3ut6vyntc-rndir.vercel.app/

brentvatne (approved):
good stuff!

Simek:
Let's ship it then! 😉

PR #691, opened 2021-08-04
Add react-native-relative-popup

ysfuu:
# Checklist

If you added a new library:


Simek (approved):
Hi @ysfuu, thank you for adding new library to the directory! 👍 

About the changes in PR, generally we prefer to stick to NPM name as a display name for the package if there is no particular reason to change name to something else. 🙂

PR #705, opened 2021-08-17
Add react-native-email-chip

arelstone:
# Why
I've been needing a component like this. If I've needed it, other developers must have needed it too

# Checklist

If you added a new library:


Simek (approved):
Hi @arelstone, thank you for adding new library to the directory.

I had to removed `nameOverride` because package with that name already exist on NPM and this would confuse the users. 

Generally, in 99,5% of cases this field should not be filled in.

PR #710, opened 2021-08-27
Added Performant React Native Font Face library

mthines:
# Why

Adds font face support for RN. The current library has performance issues, so I've made this to solve that issue :)


Simek (approved):
Hi @mthines, thank you for adding new library to the directory! 👍

PR #725, opened 2021-09-18
improve SEO, add share image

Simek:
# 📝 Why & how

This PR aims to improve Directory SEO. 

Currently all the pages have the same title and description when shared and on certain places, like discord missing `og:image` has been replaced with bare logo file.

Now every page have their own metadata, I have added few other tweaks to the existing tags, like scheme support for `theme-color` and made few small tweaks here and there.

~~In the future I would like to improve this even a bit more and include search term, when entered on the index page. 🙂~~

# ✅ Checklist

# Share image preview

~~Just a basic image in recommended ratio, but it's better than none. 😉~~

logo-og


Simek:
Ok, I have made a bit of research and it reveals to me that automatic generation of share images isn't that difficult, and there are already cloud based services offering such API. 

So I have decided to use https://cloudinary.com, unfortunately in Free tier I cannot invite anyone more in, but limits seems to be quite high and should be enough for our use-case, we will monitor this after the deploy anyway. Unless the page is shared the credit is not charged, so we can afford to experiment with this feature a bit.

I have also added the customized share images when user enters a search query, which also should be a nice addition. 🙂 

If you have any other ideas, feel free to share them, I have also few more things in mind.

# Example share previews

### Main page


### Main page with query


### Popular page

components/PageMeta.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const searchDescription = `Search results for keyword:\n"${parsedQuery}"`;
+
+  const socialImage = getShareImage({
+    title: site.title,
+    tagline: parsedQuery ? searchDescription : description,
+    cloudName: 'react-native-directory',
+    imagePublicID: 'share-bg',
+    titleColor: 'fff',
+    titleFont: 'futura',
+    taglineFont: 'futura',
```
jonsamp, line 28: Could we make the font match the site?

```
-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif
```

"Helvetica" would do as well!

jonsamp (approved):
This is a nice improvement.

I commented about the font. Also, if cloudinary only allows one account, then we should find a way to share those credentials so that we can all access it?

components/PageMeta.tsx
Simek: AFAIK Cloudinary supports only Google Fonts out of the box. 

There is an option to used the custom font too, but I'm not sure what exactly licenses the fonts used on the website have and if it's fine to upload them to service like that.

Simek:
> Also, if cloudinary only allows one account, then we should find a way to share those credentials so that we can all access it?

Good idea! It is possible to change the account data on Cloudinary, so we will need a shared email address for the directory contributors for that. 🙂

brentvatne (approved):
🌟

components/PageMeta.tsx
Simek: @jonsamp What do you think about sticking to Roboto for this PR and we can figure out better solution for that in the future?

Simek: I have also figure out that I was wrong about fonts support, not all Google Fonts are supported, there are also some other fonts, "Helvetica" also works (but some of the character looked a bit misaligned in the horizontal axis so I have decided to stick to "Roboto"). 

Unfortunately there is also no list of supported fonts and all search results lead to the support pages with vague answers or links to the custom font usage guide.

### Preview
https://res.cloudinary.com/react-native-directory/image/upload/w_1280,h_669,c_fill,q_auto,f_auto/w_760,c_fit,co_rgb:fff,g_south_west,x_480,y_372,l_text:roboto_66:React%20Native%20Directory/w_760,c_fit,co_rgb:aeb1bc,g_north_west,x_480,y_324,l_text:roboto_48:Search%20results%20for%20keyword%3A%0A%22video%22/share-bg

jonsamp (approved):
Roboto is looking great

PR #825, opened 2022-04-07
Add wix/react-native-notifications

taylorkline:
# 📝 Why & how

Actively maintained notifications library.

# ✅ Checklist


Simek (approved):
Thank you for adding another library to the directory! 🚀

> One more note, the order of packages in JSON is used for the "Recently Added" filter, so we tend to keep the order correct and only add new entries at the botom of list.

PR #853, opened 2022-07-03
update most of the dependencies to the latest version

Simek:
# 📝 Why & how

This PR updates most of the possible to update dependencies for the directory app (contains React Native, Expo and Next).

There is a small issue with newest  `eslint-config-universe` which results in `import/namespace` errors for each React Native import. I have disabled the rule for now, but I will try to investigate the problem for the root cause soon.

# ✅ Checklist


Simek:
We will need to fix Vercel deploy preview, the secret is still defined and valid. 🤷‍♂️

Issue #830, opened 2022-04-16
Supports New Architecture flag


Gustash:
With React Native 0.68 introducing an optional toggle for the New Architecture, I think it would be helpful to mention in the website if a library supports it. 

As libraries update to using the new architecture, this could be a good way for developers to get a feel for how many libraries have already moved to the new architecture. 

Maybe it would be helpful as well to mention if they support both old and new architecture, like `react-native-screens` does.

PR #870, opened 2022-08-21
basic New Arch detection and new status

Simek:
# 📝 Why & how

Refs #830

This PR add a basic New Architecture support detection based on the present of the Codegen related field in the library `package.json`.

The status in display in the secondary metadata row and links to the New Architecture intro docs on the React Native docs website, which might help the directory users to understand what this term mean, if they sees it in library box.

---

Things to consider before converting from draft:
* should we allow users to manually add this flag to the JSON entries?
* should we add filter for this status (not sure if it is a good idea, until we are sure that this basic method won't cover almost all of the cases)

# 📸 Preview

https://react-native-directory-okpe29xli-rndir.vercel.app/

### Library Box

Screenshot 2022-08-21 191622

# ✅ Checklist

### DIFF
--- a/components/Filters.tsx
+++ b/components/Filters.tsx
@@ -160,6 +160,13 @@ export const Filters = ({ query, style, basePath = '/' }: FiltersProps) => {
             title="Has TypeScript types"
             basePath={basePath}
           />
+          <ToggleLink
+            key="newArchitecture"
+            query={query}
+            paramName="newArchitecture"
+            title="Supports New Architecture"
+            basePath={basePath}
+          />
           {isMainSearch ? (
             <ToggleLink
               key="isMaintained"
--- a/components/Library/MetaData.tsx
+++ b/components/Library/MetaData.tsx
@@ -16,6 +16,7 @@ import {
   Fork,
   Code,
   TypeScript,
+  ReactLogo,
 } from '../Icons';
 import { DirectoryScore } from './DirectoryScore';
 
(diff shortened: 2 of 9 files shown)


brentvatne:
> should we allow users to manually add this flag to the JSON entries?

yeah, we should! we take a different approach to supporting fabric in expo modules and the approach in this pr won't detect it

brentvatne:
> should we add filter for this status (not sure if it is a good idea, until we are sure that this basic method won't cover almost all of the cases)

this could be useful. if we identify that something is wrong, we can adjust as we go

jonsamp (approved):
Thanks for adding this meta data!

cipolleschi (approved):

cipolleschi:
> should we allow users to manually add this flag to the JSON entries?

What do you mean here? Lib maintainer has to manually add this entry in the `package.json`, otherwise the app can't generate the code it needs.

> should we add filter for this status

Which statuses are you considering? Something like: `In development`, `iOS done, Android not done`, or other.
I guess that for the sake of the user and of simplicity, we should only use a simple boolean: whether the latest release support the new architecture or not.

tomekzaw:
> should we allow users to manually add this flag to the JSON entries?

React Native Reanimated also supports the New Architecture but doesn't have any native components or TurboModules, so the possibility to override this flag in `react-native-libraries.json` would be very appreciated.

brentvatne (approved):

PR #874, opened 2022-08-28
use Radix HoverCard instead of react-popper and popper.js

Simek:
# 📝 Why & how

We plan to utilize more Radix components in the future for filters and sort, but let's start with replacing `react-popper` and `popper.js` with `HoverCard`:
* https://www.radix-ui.com/docs/primitives/components/hover-card

I have attempted to replicate the previous appearance and behavior 1:1, unfortunately it's not possible to turn off collision detection in each axis separately, but turning it completely allow the X overflow which ruins this feature on narrower viewports/windows. So at the end for the UX I have left collision turned on, hopefully the most annoying side effect of this switch, which is preview jumping on user scroll can be disabled by forcing card to stay to initial selected side.

# ✅ Checklist

# Preview

https://react-native-directory-git-head-rndir.vercel.app/


jonsamp (approved):
Thanks for swapping this out for radix UI. this is definitely a positive direction for this directory!

PR #880, opened 2022-09-14
add rxn-units + rename rxn-keylistener

runtothedoor:
# 📝 Why & how

+ Add **rxn-units** 
📐 Viewport units (vw, vh, vmin, vmax, percentage) + useUnits hook for React, React Native and Expo.

+ Rename old package "**keylistener**" to "**rxn-input**" (Still under development but ready for android/web tests)
🎮 🖱️ ⌨️‎ ‎ Cross Platform Input Handler for React and React Native.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -8402,18 +8402,6 @@
     "web": true
   },
   {
-    "githubUrl": "https://github.com/Luffos/rxn-units",
-    "npmPkg": "rxn-units",
-    "examples": ["https://codesandbox.io/s/rxn-units-example-ox6n4h"],
-    "ios": true,
-    "android": true,
-    "web": true,
-    "expo": true,
-    "windows": true,
-    "macos": true,
-    "tvos": true
-  },
-  {
     "githubUrl": "https://github.com/wobsoriano/rn-perfect-sketch-canvas",
     "examples": ["https://github.com/wobsoriano/rn-perfect-sketch-canvas/tree/master/example"],
     "images": [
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -8397,10 +8397,22 @@
...
   },
   {
-    "githubUrl": "https://github.com/Luffos/rxn-keylistener",
+    "githubUrl": "https://github.com/Luffos/rxn-input",
     "android": true,
     "web": true
   },
+  {
+    "githubUrl": "https://github.com/Luffos/rxn-units",
+    "npmPkg": "rxn-units",
```
Simek, line 8406:
```suggestion
```

This field is not needed if the repository name is the same as package name.

react-native-libraries.json
```diff
@@ -8397,10 +8397,22 @@
...
+    "examples": ["https://codesandbox.io/s/rxn-units-example-ox6n4h"],
+    "ios": true,
+    "android": true,
+    "web": true,
+    "expo": true,
+    "windows": true,
+    "macos": true,
+    "tvos": true
```
Simek, line 8414:
```suggestion
    "windows": true,
    "macos": true,
    "tvos": true
```

For out-of-tree platforms we require an example utilizing them, no matter if library uses native code ore not, since there are a lot of minor differences between the implementation of certain RN parts, which can lead to unexpected behavior.

Simek (changes requested):
Hello @runtothedoor, thank you for adding new and updating one of the libraries already in directory! 👍 

I have left small comments on entries, also have one more related to the addition - `rxn-units` should be placed at the end of the file, because entries order is used as a source for  "Recently added" sort.

Simek (approved):
Thank you for the review updates, let's ship it! 🚀

PR #881, opened 2022-09-14
add newArchitecture to README.md

runtothedoor:
# 📝 Why & how
add `newArchitecture` (based on schema)

image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ Please follow format, fields order and indentation as seen below, skip any of th
 - #### `template`
   **(boolean)** - signify that a library is a new project template.
 - #### `newArchitecture`
-  **(boolean)** - signify that the library is using the new architecture.
+  **(boolean)** - signify that a library is using the new architecture.
 
 ---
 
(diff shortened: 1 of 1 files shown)


Simek:
I have not added the new architecture flag to the Readme on purpose, since it is only a temporary solution to overwrite the autodetection result, until we are not sure that we can cover every case, see:
* https://github.com/react-native-community/directory/pull/870
* https://github.com/microsoft/rnx-kit/issues/1863

runtothedoor:
> Nice catch, let me explain! I have not added the new architecture flag to the Readme on purpose, since it is only a temporary solution to overwrite the autodetection result, until we are not sure that we can cover every case, see:
> 
> * [basic New Arch detection and new status #870](https://github.com/react-native-community/directory/pull/870)
> * [[Feature request] Use dep-check to help migrate apps to the New Architecture microsoft/rnx-kit#1863](https://github.com/microsoft/rnx-kit/issues/1863)

Oh, I see. I thought about it but I didn't know it was already in your plans hahaha : )

runtothedoor:
Closing...

PR #882, opened 2022-09-14
add @babylonjs/react-native package + fix GITHUB_URL_PATTERN

runtothedoor:
# 📝 Why & how
Add **BabylonReactNative** related packages
https://www.babylonjs.com/reactnative/

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -8658,8 +8658,7 @@
     "githubUrl": "https://github.com/BabylonJS/BabylonReactNativeSample",
     "ios": true,
     "android": true,
-    "windows": true,
-    "template": true
+    "windows": true
   },
   {
     "githubUrl": "https://github.com/runtothedoor/rotating-cube-demo-babylon-rxn",
@@ -8668,7 +8667,6 @@
     ],
     "ios": true,
     "android": true,
-    "web": true,
-    "template": true
+    "web": true
   }
 ]
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -8663,6 +8663,9 @@
   },
   {
     "githubUrl": "https://github.com/runtothedoor/rotating-cube-demo-babylon-rxn",
+    "images": [
+      "https://raw.githubusercontent.com/runtothedoor/rotating-cube-demo-babylon-rxn/main/cube.gif"
+    ],
     "ios": true,
     "android": true,
     "web": true,
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -8644,7 +8644,7 @@
     "expo": true
   },
   {
-    "githubUrl": "https://github.com/BabylonJS/BabylonReactNative",
+    "githubUrl": "https://github.com/BabylonJS/BabylonReactNative/tree/master/Modules/%40babylonjs/react-native",
     "npmPkg": "@babylonjs/react-native",
     "examples": [
       "https://github.com/BabylonJS/BabylonReactNative/tree/master/Apps",
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -8642,5 +8642,30 @@
     "android": true,
     "web": true,
     "expo": true
+  },
+  {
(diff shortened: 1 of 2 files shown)


react-native-libraries.json
```diff
@@ -8642,5 +8642,33 @@
...
+  },
+  {
+    "githubUrl": "https://github.com/runtothedoor/rotating-cube-demo-babylon-rxn",
+    "images": [
+      "https://raw.githubusercontent.com/runtothedoor/rotating-cube-demo-babylon-rxn/main/cube.gif"
+    ],
+    "ios": true,
+    "android": true,
+    "web": true,
+    "template": true
```
Simek, line 8672: By `template` we understand not the demo apps, but the project templates not tight to the one library, rather a complex app setup for multiple use case like, for example:
* https://github.com/react-native-community/react-native-template-typescript
* https://github.com/leotm/react-native-template-new-architecture

react-native-libraries.json
```diff
@@ -8642,5 +8642,33 @@
     "android": true,
     "web": true,
     "expo": true
+  },
+  {
+    "githubUrl": "https://github.com/BabylonJS/BabylonReactNative",
```
Simek, line 8647:
```suggestion
    "githubUrl": "https://github.com/BabylonJS/BabylonReactNative/tree/master/Modules/%40babylonjs/react-native",
```

If package is a a part of monorepo, we need to specify full path to the folder which includes `package.json` of that library.

Simek:
Also, according to the referenced issue, we have a fallback to `package.json` description field, if no GitHub description is set.

runtothedoor:
> Also, according to the referenced issue, we have a fallback to `package.json` description field, if no GitHub description is set.

I see, good to know 👏

react-native-libraries.json
Simek: Sorry for not being clear enough, those two entries should be a part of `@babylonjs/react-native` examples array, instead of being a sole entries in the directory. 🙂 

Saying that, we have on our list to consider if we would like to have a demo apps directory and if yes, how we would like to tackle this topic, since mixing libraries with demo apps might cause unnecessary noise for the users, and probably with time demo apps will outnumber the libraries anyway.

There are at least few factors which we need to account for, for example, unfortunately statistically demo apps come with a lot more risk of getting outdated/unmaintained/abandoned than the libraries, especially with lively community around them, which at the end can lower the quality of RND results in time and would require more maintenance work to keep the list in shape.

runtothedoor: > Sorry for not being clear enough, those two entries should be a part of `@babylonjs/react-native` examples array, instead of being a sole entries in the directory. 🙂
> 
> Saying that, we have on our list to consider if we would like to have a demo apps directory and if yes, how we would like to tackle this topic, since mixing libraries with demo apps might cause unnecessary noise for the users, and probably with time demo apps will outnumber the libraries anyway.
> 
> There are at least few factors which we need to account for, for example, unfortunately statistically demo apps come with a lot more risk of getting outdated/unmaintained/abandoned than the libraries, especially with lively community around them, which at the end can lower the quality of RND results in time and would require more maintenance work to keep the list in shape.

Sure, no problems! 

Editing...

Simek:
> But in this case the package also doesn't include this field. Will we have any problems?

I think that in worst case this field will be empty, but this should not affect anything besides presentation.

Also, according to the CI, we need to correct schema to allow "@" in GitHub URL field, it looks like the Babylon repo is the first case which uses that naming pattern. 😅 Would you like to take care of that in this PR?

runtothedoor:
> > But in this case the package also doesn't include this field. Will we have any problems?
> 
> I think that in worst case this field will be empty, but this should not affect anything besides presentation.
> 
> Also, according to the CI, we need to correct schema to allow escaped "@" in GitHub URL field, it looks like the Babylon repo is the first case which uses that naming pattern. 😅 Would you like to take care of that in this PR?

Ok hahahaha 😄

Simek (approved):
Thank you for adding new entry and adjusting GitHub URL regexp, LGTM! 🙂

PR #886, opened 2022-09-27
Update react-native-libraries.json: Add react-native-localize platform support

redbar0n:
According to what is stated in its readme: https://github.com/zoontek/react-native-localize

# ✅ Checklist


Simek (approved):
Hello @redbar0n, thank you for entry the update! 👍 

FYI, for out-of-tree platforms we require an example containing those platform packages, which is present for this library, so it's fine to marge this in.

PR #893, opened 2022-10-16
update entry for react-native-flipper-databases

panz3r:
# 📝 Why & how

This PR updates the `react-native-flipper-databases` entry to point to the new repository

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -5766,12 +5766,12 @@
     "dev": true
   },
   {
-    "githubUrl": "https://github.com/panz3r/react-native-flipper-databases",
+    "githubUrl": "https://codeberg.org/panz3r/react-native-flipper-databases",
     "examples": [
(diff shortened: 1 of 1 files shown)


Simek:
Hello @panz3r, unfortunately at this moment of time we do not support other code hosting platforms than GitHub. 

We relay on GitHub metadata fetched via API to fill out the entry, so to properly support new host, besides updating few checks we will need to create new scripts for fetching those additional data from different provider API and alter the the build data script to accommodate workflow to new addition. 

I think that we had the similar request in the past about BitBucket, but the problem with those two platforms is that you need an account to access the source code in the first place, so the code nor examples are not publicly accessible. This creates another batch of problems, when it comes to display information and usability of such entry.

Saying that, I will discuss this topic with other maintainers and we will see what we can do to improve and support more code platforms in the directory.

PR #895, opened 2022-10-19
Rename some references from master to main

notjosh:
# 📝 Why & how

I found a reference to `master` in the `README.md`. Easy fix. Also noticed it was also referenced in the deployment yml - it seems like that should change, but happy to remove that one if we only want scheduled deploys.

💅💅💅

### DIFF
--- a/.github/workflows/deploy-to-production.yml
+++ b/.github/workflows/deploy-to-production.yml
@@ -2,7 +2,7 @@ name: Deploy to production
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


Simek (approved):
LGTM, thank you for the updates!

According to the `master` in workflow, I think that GitHub have sort of fallback mechanism for renamed base branch, since I can see that those deploys were triggered correctly, even with old branch name in them.

PR #899, opened 2022-10-26
Improve react-native-avoid-softinput examples & images

mateusz1913:
# 📝 Why & how

This PR adds missing examples and images for `react-native-avoid-softinput`, as well as it declares support for "New Architecture"

Note: `react-native-avoid-softinput` github repo is a monorepo and it turns out that the ["New Architecture" detection](https://github.com/react-native-community/directory/commit/4ae25e9fed97d8aae357a0cede3c7119d6d2a276) (btw awesome feature 🚀 ) does not support detecting inside monorepos. I also saw that 2 repos (reanimated and vector-drawable) [were added manually](https://github.com/react-native-community/directory/commit/f5cfbf23ccd498ade9efcbb15e912a04d57c6341) after detection feature landed. Technically this is undocumented, but for now, there's no other way to achieve it, so I just mirrored it for my library, let me know, if it's ok 😉 

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -6866,9 +6866,15 @@
     "android": true
   },
   {
-    "githubUrl": "https://github.com/mateusz1913/react-native-avoid-softinput",
+    "githubUrl": "https://github.com/mateusz1913/react-native-avoid-softinput/tree/master/packages/react-native-avoid-softinput",
     "examples": [
+      "https://github.com/mateusz1913/react-native-avoid-softinput/tree/master/packages/fabricMobile",
       "https://github.com/mateusz1913/react-native-avoid-softinput/tree/master/packages/mobile"
+    ],
+    "npmPkg": "react-native-avoid-softinput",
+    "images": [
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -6868,10 +6868,16 @@
   {
     "githubUrl": "https://github.com/mateusz1913/react-native-avoid-softinput",
```
Simek, line 6869:
```suggestion
    "githubUrl": "https://github.com/mateusz1913/react-native-avoid-softinput/tree/master/packages/react-native-avoid-softinput",
```

If you change the GitHub URL to directly point to the package, not the whole monorepo, the automatic detection should work. This is also the correct way to specify packages inside one, using the repo URL in that case can also affect metadata, not only new architecture detection.

Simek:
@mateusz1913 Will you perform the suggested changes or should I do that?

mateusz1913:
@Simek I'll make a changes and request a re-review in a few hours

mateusz1913:
Changes done

Simek (approved):
LGTM, thank you for the updates, let ship them! 🚀

PR #905, opened 2022-11-07
Add react-native-tooliplize library

amandeepmittal:
# 📝 Why & how

Add [react-native-tooltiplize](https://github.com/Stringsaeed/react-native-tooltiplize) which is a new universal tooltip library.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -8867,7 +8867,6 @@
   },
   {
     "githubUrl": "https://github.com/Stringsaeed/react-native-tooltiplize",
-    "npmPkg": "https://www.npmjs.com/package/react-native-tooltiplize",
     "examples": ["https://github.com/Stringsaeed/react-native-tooltiplize/tree/main/example"],
     "images": [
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -8864,5 +8864,20 @@
     ],
     "ios": true,
     "android": true
+  },
+  {
+    "githubUrl": "https://github.com/Stringsaeed/react-native-tooltiplize",
+    "npmPkg": "https://www.npmjs.com/package/react-native-tooltiplize",
```
Simek, line 8870:
```suggestion
```

This field only should contain npm package name to validate correctly, but it is not needed when GitHub repository name matches package name.

Simek (commented):
Thank you for adding new library to the directory! 👍 

Left a suggestion which should fix the CI, besides that LGTM.

react-native-libraries.json
amandeepmittal: Thanks for the update!

Simek (approved):
🚀

PR #911, opened 2022-11-25
Update henninghall/react-native-date-picker expo support

l1qu1d:
# 📝 Why & how
Updating https://github.com/henninghall/react-native-date-picker to set `expo` support to `false`.
While the project is expo compatible, it doesn't fully support **managed workflow** because it cannot run in `Expo Go`.

Excerpt from the repo: "_This library can't be used in the "Expo Go" app because it [requires custom native code](https://docs.expo.dev/workflow/customizing/)._"

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -23,9 +23,7 @@
     "android": true,
     "web": true,
     "expo": true,
-    "examples": [
-      "https://github.com/Iconscout/react-native-unicons/tree/master/example"
-    ],
+    "examples": ["https://github.com/Iconscout/react-native-unicons/tree/master/example"],
     "npmPkg": "@iconscout/react-native-unicons"
   },
   {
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -2413,7 +2413,7 @@
     ],
     "ios": true,
     "android": true,
-    "expo": true
+    "expo": false
```
Simek, line 2416:
```suggestion
```

We can just removed the falsy property.

Simek (commented):
Hello, thank you for the update @l1qu1d! 👍 I have left one, minor suggestion to address before we ship that PR.

react-native-libraries.json
l1qu1d: Thanks! I didn't look at other objects as a basis for conformity and will do so in the future.

PR #916, opened 2022-12-10
add sort direction to directory API and website

Simek:
# 📝 Why & how

Currently we always use the descending order when presenting results. Let's add an ability to change the order via API query param ans support that feature on the directory website.

I have also added a logic to skip default sort related values from the query on the website and fixed the sorting-related types.

Additionally, I have fixed few typos and made formatting changes in project Readme.

# ✅ Checklist

# 📷 Preview

demo

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,15 +23,15 @@
 - Add it at the end of `react-native-libraries.json` file.
 - Submit a PR.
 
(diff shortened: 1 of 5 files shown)


jonsamp (approved):
this is a nice to have feature and it looks like it works

PR #930, opened 2023-01-05

arpit-bhalla:
# 📝 Why & how

React Native Elements v4 is renamed to `@rneui/themed`

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -650,7 +650,7 @@
     "unmaintained": false,
     "dev": false,
     "template": false
-  }
+  },
   {
     "githubUrl": "https://github.com/satya164/react-native-tab-view",
     "ios": true,
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -635,6 +635,23 @@
     "expo": true
   },
   {
+    "githubUrl": "https://github.com/react-native-elements/react-native-elements",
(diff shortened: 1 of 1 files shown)


Simek (commented):
Hello @arpit-bhalla, React Native Elements is already listed in the directory, and CI fails because you are trying to add the duplicate entry for it:
* https://github.com/react-native-community/directory/blob/main/react-native-libraries.json#L630-L636

I can see that the project structure has changed, it is now a mono repo with the separated and renamed packages in there. In this case we probably should update the current entry to point to `@rneui/base` and then add `@rneui/themed` as an additional package.

Simek:
If the repository is the same for the v3 and v4 it is not possible to have multiple, separate entries ATM.

What's the motivation for this split?

Simek:
Sorry for not being clear enough, I have meant the separation of versions, so v3 and v4. 

Why do you think it will be beneficial to have both versions at the same time? Is something wrong with making the current entry `@rneui/base`?

Simek:
Looks like we have the entry for `rneui` now:

* https://github.com/react-native-community/directory/blob/0dc3501133776139f704839b258a66d88fe92e08/react-native-libraries.json#L686-L692

Please create a new PR if the existing entry need an update.

PR #936, opened 2023-01-11
Update react-native-system-navigation-bar

kadiraydinli:
# 📝 Why & how

Update `react-native-system-navigation-bar`'s screenshots and expo support.

# ✅ Checklist


Simek (approved):
LGTM, thank you for the update! 👍 

Just to clarify, every package can be used with Expo while using development client (https://docs.expo.dev/development/introduction), but indeed, the flag represents Expo Go app support and this package will not work out of the box with this store app.

PR #941, opened 2023-01-17
Add newArchitecture to README

sAleksovski:
# 📝 Why & how

There is a `newArchitecture` property defined in `react-native-libraries.schema.json`, and is already used by some libraries, but it was not documented in `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,8 @@
   "tvos": false,
   "unmaintained": false,
   "dev": false,
-  "template": false
+  "template": false,
+  "newArchitecture": false
 }
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,10 +102,12 @@
   **(boolean)** - signify that a library is a development tool or is only a part of development process.
 - #### `template`
   **(boolean)** - signify that a library is a new project template.
+- #### `newArchitecture`
+  **(boolean)** - signify that a library supports the new architecture
```
Simek, line 106:
```suggestion
  **(boolean)** - signify that a library supports the new architecture

  > Set this tag only when automatic architecture detection fails for your package, despite it supports the new architecture.
```

Let's mention briefly the tag/field purpose.

Simek:
For the context: The reason is that this field is only a temporary solution, workaround for the cases when automatic detection do not work, but there are plans in core team to standardize and formalize the `package.json` fields for RN, which in the future allows us to fully relay on automatic detection and remove this field.

But I guess we can list it in Readme, for now. I have made a suggestion which adds the explanation of the purpose of this field.

sAleksovski:
> For the context: The reason is that this field is only a temporary solution, workaround for the cases when automatic detection do not work, but there are plans in core team to standardize and formalize the `package.json` fields for RN, which in the future allows us to fully relay on automatic detection and remove this field.
> 
> But I guess we can list it in Readme, for now. I have made a suggestion which adds the explanation of the purpose of this field.

I didn't know that, just saw that it was missing in the README.
You can decline the PR if you think it's better for it not to be included.

Simek (approved):
Let's add this field for now, thank you for the review update, LGTM! 👍

PR #942, opened 2023-01-17
Add @rest-hooks/react

ntucker:
# 📝 Why & how

Adds the [@rest-hooks/react](https://www.npmjs.com/package/@rest-hooks/react) library aka [Rest Hooks](https://resthooks.io/)

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -9247,5 +9247,16 @@
     "ios": true,
     "android": true,
     "dev": true
+  },
+  {
+    "githubUrl": "https://github.com/coinbase/rest-hooks",
+    "npmPkg": "@rest-hooks/react",
+    "examples": [
+      "https://github.com/coinbase/rest-hooks/tree/master/examples",
+      "https://resthooks.io/demos"
+    ],
+    "ios": true,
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -9247,5 +9247,19 @@
...
+    "examples": [
+      "https://github.com/coinbase/rest-hooks/tree/master/examples",
+      "https://resthooks.io/demos"
+    ],
+    "ios": true,
+    "android": true,
+    "web": true,
+    "expo": true,
+    "windows": true,
+    "macos": true
```
Simek, line 9263: Hello @ntucker, thank you for the new addition! 👍 

For the out-of-tree platforms we require an example, since there are difference in APIs and behavior when comparing to the core RN, and I was not able to find any for macOS or Windows.

Also, the `expo` flag signifies the support for Expo Go, so the managed flow, where user cannot use/overwrite `AppRegistry` in a way which is described in package docs, so I would suggest to drop that flag too.

Simek (approved):
Thank you for the updates, LGTM! 🚀

PR #949, opened 2023-02-02
Add module react-native-idfa-aaid

edodusi:
# 📝 Why & how

This adds the library https://github.com/sparkfabrik/sparkfabrik-react-native-idfa-aaid

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -9299,7 +9299,6 @@
     "githubUrl": "https://github.com/sparkfabrik/sparkfabrik-react-native-idfa-aaid",
     "npmPkg": "@sparkfabrik/react-native-idfa-aaid",
     "ios": true,
-    "android": true,
-    "expo": true
+    "android": true
   }
 ]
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -9294,5 +9294,12 @@
       "https://github.com/Splicer97/react-native-osmdroid/tree/main/example"
     ],
     "android": true
+  },
+  {
+    "githubUrl": "https://github.com/sparkfabrik/sparkfabrik-react-native-idfa-aaid",
+    "npmPkg": "@sparkfabrik/react-native-idfa-aaid",
+    "ios": true,
+    "android": true,
+    "expo": true
```
Simek, line 9303:
```suggestion
    "android": true
```

Expo flag means that the library will work with Expo Go app (every library can work with `dev-client`s), and after the quick glance at the package source, it probably won't work. It would be nice to confirm that, or just remove the `expo` flag for now.

edodusi: @Simek sorry I didn't check, I was thinking just "it will work with an Expo app" which is true because I use it also in an Expo app but now I get it, changed the line!

Simek (approved):
Thank you for the update, LGTM! 👍

PR #959, opened 2023-03-09
Update react-native-app-auth

keithluchtel:
# 📝 Why & how

Updates react-native-app-auth to note that it does not support Expo Go

# ✅ Checklist


Simek (approved):
Thank you for the update, LGTM! 👍

PR #970, opened 2023-04-02
Adding bambuser broadcaster to the directory

fansilva1991:
# 📝 Why & how

Adding the [react-native-bambuser-broadcaster](https://github.com/bambuser/react-native-bambuser-broadcaster) to the directory.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -9518,5 +9518,19 @@
     "ios": true,
     "android": true,
     "dev": true
+  },
+  {
+    "githubUrl": "https://github.com/bambuser/react-native-bambuser-broadcaster",
+    "ios": true,
+    "android": true,
+    "web": false,
+    "expo": false,
+    "windows": false,
+    "macos": false,
+    "tvos": false,
+    "unmaintained": true,
+    "dev": false,
+    "template": false,
+    "newArchitecture": false
   }
 ]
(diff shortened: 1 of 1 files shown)


fansilva1991:
Hi @Simek, this library is still compatible with the latest version of React Native; 

I'm using it with version ```0.71.3```

Sadly, there is no alternative for this library since this has not been updated by the vendor ( Bambuser );

I think we should add this since this is the only alternative available, and developers can even use it to sign to their teams that the library is unmaintained, making them sure about the risks of using it.

But I will understand if we decide not to add this to the directory :)

Thanks for the response

react-native-libraries.json
```diff
@@ -9518,5 +9518,19 @@
...
+    "android": true,
+    "web": false,
+    "expo": false,
+    "windows": false,
+    "macos": false,
+    "tvos": false,
+    "unmaintained": true,
+    "dev": false,
+    "template": false,
+    "newArchitecture": false
```
Simek, line 9534: Thank you for the explanation and checks! 👍  If the library works with the latest version of React Native it is fine to add it, besides the `unmaintained` flag I think, so LGTM.

Let's just clean up the entry, before we merge that in:
```suggestion
    "android": true,
    "unmaintained": true
```

fansilva1991: It is done as suggested :)

Simek (approved):
Thank you for the updates! 🚀

PR #975, opened 2023-04-28
Adding react-native-use-list

dozsolti:
# 📝 Why & how

Added my `react-native-use-list` library which helps you with common lists problems.
https://github.com/dozsolti/react-native-use-list

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -9548,8 +9548,6 @@
   },
   {
     "githubUrl": "https://github.com/dozsolti/react-native-use-list",
-    "npmPkg": "react-native-use-list",
-    "nameOverride": "React Native useList",
     "examples": [
       "https://github.com/dozsolti/react-native-use-list/blob/main/example/examples/FlatList/Pagination.tsx"
     ],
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -9545,5 +9545,17 @@
     ],
     "ios": true,
     "android": true
+  },
+  {
+    "githubUrl": "https://github.com/dozsolti/react-native-use-list",
+    "npmPkg": "react-native-use-list",
+    "nameOverride": "React Native useList",
```
Simek, line 9552: Hello @dozsolti, thank you for the contribution to the directory! 👍 

```suggestion
```

`npmPkg` is not needed since the GitHub repository name matches the package name, also when it comes for `nameOverride`, common practice is to stick to the package name, unless there is a good reason not to do it.

react-native-libraries.json
```diff
@@ -9545,5 +9545,17 @@
...
+    "examples": [
+      "https://github.com/dozsolti/react-native-use-list/blob/main/example/examples/FlatList/Pagination.tsx"
```
Simek, line 9554:
```suggestion
      "https://github.com/dozsolti/react-native-use-list/tree/main/example"
```

Usually, we link users to the example app root.

Simek (approved):
Thank you for the updates, LGTM! 👍

PR #1052, opened 2023-10-05
Added react-native-orientation-manager

KroosX4V:

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -10050,5 +10050,12 @@
     "ios": true,
     "android": true,
     "expo": true
+  },
+  {
+    "githubUrl": "https://github.com/KroosX4V/react-native-orientation-manager",
+    "examples": ["https://github.com/KroosX4V/react-native-orientation-manager/tree/main/example"],
+    "ios": true,
+    "android": true,
+    "windows": true
   }
 ]


react-native-libraries.json
```diff
@@ -10016,5 +10016,12 @@
     ],
     "ios": true,
     "android": true
+  },
+  {
+    "githubUrl": "https://github.com/KroosX4V/react-native-orientation-manager",
+    "npmPkg": "react-native-orientation-manager",
```
Simek, line 10022:
```suggestion
```

This field can be skipped, since the repository name matches package name, it's a field dedicated mostly for scoped packages.

react-native-libraries.json
```diff
@@ -10016,5 +10016,12 @@
...
+    "ios": true,
+    "android": true,
+    "windows": true
```
Simek, line 10025: For the out-of-tree platforms support (like Windows) we usually require an example app, since there were several false-positives compatibility marks in the past. 

To do not block this PR, we can ship it with mobile platforms, and update the Windows support when you will have the example available.

Simek (commented):
Hey @KroosX4V, thank you for adding new library to the directory! 👍 

I have left two suggestions above, with those addressed we should be able to ship the changes.

KroosX4V:
Hey @Simek, thank you for your reply.

I've resolved the first issue. For the example app, once it's ready I'm going to make a commit to this pull request with which the example is included. However, I'm not sure if this a valid workflow. If it's not, you can dismiss this PR. I'll create another one once the example app is ready.

Simek:
We can leave this PR open until the example is ready to share, or we can remove the `windows: true` flag for now, and ship addition as is, and then, in the future you can make an another PR adding the flag back with the example app URL.

KroosX4V:
Hey @Simek, I've added an example app. Please review the changes.

PR #1057, opened 2023-10-17
Update `realm` in `react-native-libraries.json`

elle-j:
# 📝 Why & how
Updates the `realm` library in `react-native-libraries.json`.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -6936,8 +6936,7 @@
       "https://www.mongodb.com/docs/atlas/app-services/tutorial/react-native/"
     ],
     "ios": true,
-    "android": true,
-    "expo": true
+    "android": true
   },
   {
     "githubUrl": "https://github.com/georstat/react-native-image-cache",
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -6927,14 +6927,17 @@
     "android": true
   },
   {
-    "githubUrl": "https://github.com/realm/realm-js",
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -6927,14 +6927,17 @@
...
-      "https://github.com/mongodb-university/realm-tutorial-react-native"
+      "https://github.com/realm/realm-js/tree/main/examples/rn-todo-list",
+      "https://github.com/realm/realm-js/tree/main/examples/rn-multiple-realms",
+      "https://github.com/realm/realm-js/tree/main/examples/rn-connection-and-error",
+      "https://www.mongodb.com/docs/atlas/app-services/tutorial/react-native/"
     ],
     "ios": true,
-    "android": true
+    "android": true,
+    "expo": true
```
Simek, line 6940: Hello @elle-j, thank you for the updates! 👍

```suggestion
    "android": true
```

The `expo` flag signifies the support Expo Go client, and since the library uses native code, it won't be supported (until pre-bundled). Every RN package can work with Expo when using Development Client, and it looks like it's the way recommended in template:
* https://github.com/realm/realm-js/tree/main/templates/expo-template#readme

elle-j: Thanks for the clarification! Wasn't entirely clear in the README that it referred to Expo Go. I've pushed a commit removing the `expo` entry 👍

Simek: We will update the description, thanks for the feedback! 👌

Simek (approved):
Thank you for the updates! 🚀

PR #1077, opened 2023-11-25
Add zeego

wodin:
# 📝 Why & how

Add zeego to `react-native-libraries.json`
Zeego is in a monorepo with docs, but it is the only package under `packages`. I added the link to the package as the GitHub URL. Wasn't sure if it would be better to add the link to the whole repository.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -10218,5 +10218,17 @@
     "npmPkg": "@azesmway/react-native-unity",
     "ios": true,
     "android": true
+  },
+  {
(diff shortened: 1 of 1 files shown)


Simek (approved):
Hi @wodin, thanks for adding new entry to the directory! 🙏 

Also, thanks for the context on repository URL. Can confirm that this looks correct, since the `package.json` for the package is in the subdirectory you pointed to.

Let's ship it!

PR #1101, opened 2024-01-06
Added react-native-full-responsive

Mhp23:
# 📝 Why & how

Added react-native-full-responsive

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -10442,7 +10442,6 @@
   },
   {
     "githubUrl": "https://github.com/Mhp23/react-native-full-responsive",
-    "npmPkg": "react-native-full-responsive",
     "examples": [
       "https://github.com/Mhp23/react-native-full-responsive/tree/main/example"
     ],
@@ -10455,9 +10454,6 @@
     "android": true,
     "web": true,
     "expo": true,
-    "windows": true,
-    "macos": true,
-    "tvos": true,
     "newArchitecture": true
   }
 ]
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -10439,5 +10439,25 @@
       "https://raw.githubusercontent.com/Richou/react-native-android-location-enabler/main/assets/android-location-enabler.gif"
     ],
     "android": true
+  },
+  {
+    "githubUrl": "https://github.com/Mhp23/react-native-full-responsive",
+    "npmPkg": "react-native-full-responsive",
+    "examples": [
+      "https://github.com/Mhp23/react-native-full-responsive/tree/main/example"
+    ],
+    "images": [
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -10439,5 +10439,25 @@
       "https://raw.githubusercontent.com/Richou/react-native-android-location-enabler/main/assets/android-location-enabler.gif"
     ],
     "android": true
+  },
+  {
+    "githubUrl": "https://github.com/Mhp23/react-native-full-responsive",
+    "npmPkg": "react-native-full-responsive",
```
Simek, line 10445:
```suggestion
```

This field is redundant, it's mostly dedicated for scoped apckages.

react-native-libraries.json
```diff
@@ -10439,5 +10439,25 @@
...
+      "https://github.com/Mhp23/react-native-full-responsive/raw/main/sim-2x.png",
+      "https://github.com/Mhp23/react-native-full-responsive/raw/main/sim-3x.png"
+    ],
+    "ios": true,
+    "android": true,
+    "web": true,
+    "expo": true,
+    "windows": true,
+    "macos": true,
+    "tvos": true,
```
Simek, line 10460:
```suggestion
    "windows": true,
    "macos": true,
    "tvos": true,
```

For out-of-tree platforms we require an example app confirming support:
* https://github.com/react-native-community/directory?tab=readme-ov-file#%EF%B8%8F-out-of-tree-platforms

Looks like the example app targets only build in platforms and web. To do not block addition, the flags can be re-added later in time.

Simek (commented):
Hello @Mhp23, thank you for adding a new entry! 👍 

Left two suggestions below.

react-native-libraries.json
Mhp23: Hi @Simek, you're welcome,
The package needs to use `Dimensions`, `useWindowDimensions` and `PixelRatio` as utility built-in React Native and there are no additional native modules to linking or different usage approaches. Does it's an essential to add examples for those based this case?

Simek: Unfortunately yes, since there could be a difference between implementation and result of build-in methods in out-of-tree platforms. Making an exception will cause more libraries authors to want to add those flags in similar case, then it will extrapolate to JS only packages (no native code), and as we know from the past, this not guarantee full/correct support, even when it seems like it should.

We have some troubles with compatibility claims in the past, when there was no restrictions, that's why we decided to add this validation step, when declaring out-of-tree platform support, hope you can understand the reasoning behind this restriction.

Mhp23: Okay, understood. So, based on the explanations, the flags for Android, iOS, and web should be enabled for now., once examples for these platforms have been added, I will then submit another PR.

Simek (approved):
Thanks for the updates, LGTM! 👍

PR #1107, opened 2024-01-16
Add unmaintained to react-native-billing

amandeepmittal:
# 📝 Why & how

Recently, was going through the directory to find what libraries are available for in app purchases in our ecosystem. Found that [`react-native-billing`](https://github.com/idehub/react-native-billing) is no longer maintained.

Updated `react-native-libraries.json` to reflect the same.


Simek (approved):
Nice catch, thanks for the update! 👍

PR #1112, opened 2024-01-25
feat: update unistyles metadata

jpudysz:
# 📝 Why & how

Based on the last changes and renaming `expo` flag to `expoGo` I want to update unistyles metadata.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -10133,7 +10133,6 @@
     ],
     "ios": true,
     "android": true,
-    "expoGo": false,
     "web": true,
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -10133,7 +10133,7 @@
     ],
     "ios": true,
     "android": true,
-    "expoGo": true,
+    "expoGo": false,
```
Simek, line 10136:
```suggestion
```

Nit, we can just skip false flags.

PR #1114, opened 2024-01-26
chore: update react-native-image-marker info in react-native-librarie…

JimmyDaddy:
…s.json

# 📝 Why & how

Update `react-nnative-image-marker`'s  examples and expo supportability 

# ✅ Checklist

- [x] Documented in this PR how you fixed or created the feature. [doc](https://github.com/JimmyDaddy/react-native-image-marker?tab=readme-ov-file#expo)

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -2345,7 +2345,6 @@
     "githubUrl": "https://github.com/JimmyDaddy/react-native-image-marker",
     "ios": true,
     "android": true,
-    "expoGo": true,
     "examples": [
       "https://github.com/JimmyDaddy/react-native-image-marker/tree/master/example",
       "https://github.com/JimmyDaddy/react-native-image-marker/tree/master/expo-example"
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -2345,7 +2345,11 @@
     "githubUrl": "https://github.com/JimmyDaddy/react-native-image-marker",
     "ios": true,
     "android": true,
-    "examples": ["https://github.com/JimmyDaddy/react-native-image-marker/tree/master/example"],
+    "expoGo": true,
```
Simek, line 2348: Hello @JimmyDaddy, thanks you for the update! 👍 

Also thank you for adding better Expo support, unfortunately your library will not work with Expo Go due to native files used. Config plugins are a part of `prebuild`, and their goal is to automate CNG processes for users. Expo Go ships with native files already bundled, and they cannot be extended.

We talk about it a bit more in the [Readme file](https://github.com/react-native-community/directory?tab=readme-ov-file#expogo), where we also explain why there is no "Expo" flag.

```suggestion
```

JimmyDaddy: I set it up according to react-native-vision-camera in libraries.json .😂

Simek (approved):
Let's ship it! 🚀

PR #1139, opened 2024-03-02
mark as unmaintained EkoLabs/react-native-background-downloader

kesha-antonov:
Hi

This repo is unmaintained for 2 years already

https://github.com/EkoLabs/react-native-background-downloader


Simek (approved):
This change make sense, thanks for the update! 🙏

PR #1147, opened 2024-03-16
Updated expo-drag-drop-content-view data

AlirezaHadjar:
# 📝 Why & how

# ✅ Checklist

- Changed example and gif paths due to migrating to monorepo

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -9880,7 +9880,7 @@
     "expoGo": true
   },
   {
-    "githubUrl": "https://github.com/AlirezaHadjar/expo-drag-drop-content-view",
+    "githubUrl": "https://github.com/AlirezaHadjar/expo-drag-drop-content-view/tree/main/packages/expo-drag-drop-content-view",
     "examples": ["https://github.com/AlirezaHadjar/expo-drag-drop-content-view/tree/main/apps/example"],
     "images": [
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -9881,10 +9881,12 @@
   },
   {
     "githubUrl": "https://github.com/AlirezaHadjar/expo-drag-drop-content-view",
```
Simek, line 9883:
```suggestion
    "githubUrl": "https://github.com/AlirezaHadjar/expo-drag-drop-content-view/tree/main/packages/expo-drag-drop-content-view",
```

The package URL also need to be updated, to correctly extract the package data after monorepo reorganization.

Simek:
Not strictly related to the changes in this PR, looks like you are using the links in package Readme which leads to Expo API reference, thus they results in 404.

react-native-libraries.json
AlirezaHadjar: Correct, thanks for the tip🙂

AlirezaHadjar:
> Not strictly related to the changes in this PR, looks like you are using the links in package Readme which leads to Expo API reference, thus they results in 404.
> 
> Screenshot 2024-03-18 105145

Thanks for the heads up. I just fixed the broken links as well.

Simek (approved):
Thanks for the updates! 👍

PR #1195, opened 2024-07-01
fix(libraries): rn-voice/voice does not support Expo Go

frankcalise:
# 📝 Why & how
- Addresses a bug in the listing for `@react-native-voice/voice`, where it shows it supports Expo Go. As per the [README documentation](https://github.com/react-native-voice/voice?tab=readme-ov-file#prebuild-plugin), it does not.
- Simply changes the `expoGo` boolean to false

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -3228,7 +3228,6 @@
     "githubUrl": "https://github.com/react-native-voice/voice",
     "ios": true,
     "android": true,
-    "expoGo": false,
     "npmPkg": "@react-native-voice/voice"
   },
   {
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -3228,7 +3228,7 @@
     "githubUrl": "https://github.com/react-native-voice/voice",
     "ios": true,
     "android": true,
-    "expoGo": true,
+    "expoGo": false,
```
Simek, line 3231:
```suggestion
```

Hello @frankcalise, nice catch, thanks for the update! 👍 

In that case we can also just remove the `expoGo` field from the package entry.

Simek (approved):

PR #1198, opened 2024-07-17
Update Bottom Sheet

ThusSpokeNomad:
# 📝 Why & how

Update bottom sheet (as requested by @brentvatne over here: https://github.com/reactwg/react-native-new-architecture/discussions/167#discussioncomment-10056394)

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -24,21 +24,21 @@
     "android": true,
     "web": true,
     "expoGo": true,
-    "examples": ["https://github.com/Iconscout/react-native-unicons/tree/master/example"],
+    "examples": [ "https://github.com/Iconscout/react-native-unicons/tree/master/example" ],
     "npmPkg": "@iconscout/react-native-unicons"
   },
   {
     "githubUrl": "https://github.com/imgly/pesdk-react-native",
     "ios": true,
     "android": true,
-    "examples": ["https://github.com/imgly/pesdk-react-native-demo"],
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -11224,15 +11224,24 @@
...
     "ios": true
+  },
+  {
+    "githubUrl": "https://github.com/gorhom/react-native-bottom-sheet",
+    "examples": [ "https://github.com/gorhom/react-native-bottom-sheet/tree/master/example" ],
+    "images": [ "https://raw.githubusercontent.com/gorhom/react-native-bottom-sheet/master/preview.gif" ],
+    "npmPkg": "@gorhom/bottom-sheet",
+    "android": true,
+    "ios": true,
+    "newArchitecture": false
```
ThusSpokeNomad, line 11245: ^ this is the new entry.

Simek:
Hey @ThusSpokeNomad, looks like something lead to the reformatting in the whole data file. Usually, we would need to revert those changes, but in this case, it looks like we can close the PR, since given library is already present in the directory:
* https://reactnative.directory/?search=%40gorhom%2Fbottom-sheet

We have a deduplicate mechanism, which should fail the CI, but it did not work in this case for some reason, going to investigate the checks.

ThusSpokeNomad:
hey @Simek nice catch, I didn't see that!
let's consider this as an update then?
I also reverted formatting changes too.

Simek (approved):
Sounds good, thanks for the updates! 👍

PR #1199, opened 2024-07-17
Add hidden filter for unmaintained libs

brentvatne:
# 📝 Why & how

It can occasionally be useful to look through a list of *only unmaintained libraries*. This makes it possible by adding the `isMaintained=false` query param

### DIFF
--- a/util/search.js
+++ b/util/search.js
@@ -117,7 +117,11 @@ export const handleFilterLibraries = ({
       return false;
     }
 
-    if (isMaintained && library.unmaintained) {
+    if (isMaintained === 'false' && !library.unmaintained) {
+      return false;
+    }
+
+    if (isMaintained === 'true' && library.unmaintained) {
       return false;
     }
 


Simek:
Thanks! In current state this will only add the functionality to the API, if you want to see the filter based on that on the website you also will need to edit the following component:
* https://github.com/react-native-community/directory/blob/main/components/Filters/index.tsx

brentvatne:
> Thanks! In current state this will only add the functionality to the API, if you want to see the filter based on that on the website you also will need to edit the following component:
> 
> * https://github.com/react-native-community/directory/blob/main/components/Filters/index.tsx

I'm ok with just filtering it based off of the query param for now! Hesitant to add more options to the UI.. although it might be interesting to expose kind of multi-state checkboxes, eg: true / false / null represented with `[✓]` `[x]` `[ ]` - that can be for a future pr

PR #1201, opened 2024-07-17
Support filtering by not supported and untested with new arch

brentvatne:
# 📝 Why & how

I find this useful to have, maybe we expose it in the UI eventually. Similar idea to #1199.

I pulled new arch status detection into a util to reuse in search filters.


brentvatne:
works well on preview: https://react-native-directory-8sojsrrya-rndir.vercel.app/

so I'll land it

PR #1203, opened 2024-07-18
Add newArchitectureNote and alternatives fields, update data and UI

brentvatne:
# 📝 Why & how

- I wanted to split out the alternative libraries from `unmaintained`, since we can use it for both unmaintained and new arch options - and perhaps further cases in the future. I also wanted to support listing multiple alternatives.
- I wanted to split out the new arch note from `newArchitecture` because I want to add notes regardless of the supported state (true/false/untested).
- I updated the data and UI accordingly.

Check it out on the preview.

# ✅ Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,16 +123,24 @@
 #### 🏷️ Tags
 
 - #### `unmaintained`
-  **(boolean|string)** - signify that a library is no longer maintained. Can also pass the package name string to point out the recommended alternative.
+  **(boolean)** - signify that a library is no longer maintained. Provide alternative libraries with the `alternatives` field.
 - #### `dev`
   **(boolean)** - signify that a library is a development tool or is only a part of development process.
 - #### `template`
   **(boolean)** - signify that a library is a new project template.
 - #### `newArchitecture`
 
-  **(boolean)** - signify that a library supports, or not, the new architecture. Skipping the field will result in "untested" status, unless automatic support detection returned a result. 
+  **(boolean)** - signify that a library supports, or not, the New Architecture. Skipping the field will result in "untested" status, unless automatic support detection returned a result.  Provide additional context with the `newArchitectureNote` field.
 
-  > Set this tag only when automatic architecture detection fails for your package, despite it supports the new architecture.
+  > Set this tag only when automatic architecture detection fails for your package, despite it supports the New Architecture.
+
+### 📝 Additional context for tags
+
+- #### `newArchitectureNote`
+  **(string)** - provide a note for the New Architecture support status, if a boolean `"true"` or `"false"` is not sufficient to describe the state of New Architecture support.
+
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -123,16 +123,24 @@
...
-  > Set this tag only when automatic architecture detection fails for your package, despite it supports the new architecture.
+  > Set this tag only when automatic architecture detection fails for your package, despite it supports the New Architecture.
+
+### 📝 Additional context for tags
+
+- #### `newArchitectureNote`
+  **(string)** - provide a note for the New Architecture support status, if a boolean `"true"` or `"false"` is not sufficient to describe the state of New Architecture support.
+
+- #### `alternatives`
+  **(array of strings)** - provide a list of alternatives to the library. eg: `["expo-camera", "react-native-vision-camera"]`. This is used to provide a list of alternatives to a library if it is unmaintained or does not suppport the New Architecture.
```
Simek, line 143:
```suggestion
  **(array of strings)** - provide a list of alternatives to the library. eg: `["expo-camera", "react-native-vision-camera"]`. This is used to provide a list of alternatives to a library if it is unmaintained or does not support the New Architecture.
```

Typo fix, also are you sure that the description reflects the logic in code?

README.md
```diff
@@ -123,16 +123,24 @@
 #### 🏷️ Tags
 
 - #### `unmaintained`
-  **(boolean|string)** - signify that a library is no longer maintained. Can also pass the package name string to point out the recommended alternative.
+  **(boolean)** - signify that a library is no longer maintained. Provide alternative libraries with the `alternatives` field.
```
Simek, line 126:
```suggestion
  **(boolean)** - signify that a library is no longer maintained. You can provide alternative or replacement libraries with the `alternatives` field, if needed.
```

Let's make it clear, that those additional fields are not required, when you set a flag.

README.md
```diff
@@ -123,16 +123,24 @@
...
 - #### `dev`
   **(boolean)** - signify that a library is a development tool or is only a part of development process.
 - #### `template`
   **(boolean)** - signify that a library is a new project template.
 - #### `newArchitecture`
 
-  **(boolean)** - signify that a library supports, or not, the new architecture. Skipping the field will result in "untested" status, unless automatic support detection returned a result. 
+  **(boolean)** - signify that a library supports, or not, the New Architecture. Skipping the field will result in "untested" status, unless automatic support detection returned a result.  Provide additional context with the `newArchitectureNote` field.
```
Simek, line 133:
```suggestion
  **(boolean)** - signify that a library supports, or not, the New Architecture. Skipping the field will result in "untested" status, unless automatic support detection returned a result.  You can provide additional context with the `newArchitectureNote` field, if needed.
```

Simek (approved):

PR #1204, opened 2024-07-18
Fix new arch label display for alternatives and update more library data

brentvatne:
# 📝 Why & how

- Fixes an issue pointed out by @Simek where we only displayed alternatives when there was a note.
- Updates more library metadata


Simek (commented):
With those changes we show the information twice for the deprecated libraries, let's only show the alternatives in. tooltip, when library is not marked as `unmaintained`.

Screenshot 2024-07-19 at 09 13 22

PR #1207, opened 2024-07-20
chore: update image and example for react-native-turbo-image

duguyihou:
# 📝 Why & how

update image and example for react-native-turbo-image

# ✅ Checklist


Simek (approved):
Thanks for the update! 👍

PR #1232, opened 2024-08-07
Update react-native-libraries.json with some packages that support ne…

gregcorp:
# 📝 Why & how
Hi, I updated expo-share-intent and react-native-app-intro-slider as they support new architechture.

# ✅ Checklist


Simek (approved):
Thanks for the updates! 👍

PR #1252, opened 2024-08-25
update: react-native-quick-crypto

boorad:
# 📝 Why & how
I wasn't seeing the `react-native-quick-crypto` library in the directory while searching for `crypto` or `quick`.  So I'm wondering if that was because it was missing the `npmPkg` field.

I added that field and a couple of others, as `v1.0.0` and higher supports new architecture.

If this doesn't fix the library not appearing in search or the directory in general, can a maintainer weigh in on what needs to change?  🙏 

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -10990,9 +10990,12 @@
   },
   {
     "githubUrl": "https://github.com/margelo/react-native-quick-crypto",
+    "npmPkg": "react-native-quick-crypto",
     "examples": ["https://github.com/margelo/react-native-quick-crypto/tree/main/example"],
     "ios": true,
-    "android": true
+    "android": true,
+    "newArchitecture": true,
+    "newArchitectureNote": "use v1.0.0 or higher for new architecture support"
   },
   {
     "githubUrl": "https://github.com/vokhuyetOz/react-native-messy",
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -10989,13 +10989,13 @@
     "expoGo": true
   },
   {
-    "githubUrl": "https://github.com/margelo/react-native-quick-crypto",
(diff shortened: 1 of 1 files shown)


Simek:
Hey @boorad, thanks for the updates! 👍 

> I wasn't seeing the react-native-quick-crypto library in the directory while searching for crypto or quick. So I'm wondering if that was because it was missing the npmPkg field.

We use the npm keywords and GitHub topics to enhance the search, and `npmPkg` is autogenerated, if not present. I have looked at the repository, and the problem is that we point to the monorepo, and not to the package, which malforms the entry. We need to change the `githubUrl` to, so the correct data is extracted:
* https://github.com/margelo/react-native-quick-crypto/tree/main/packages/react-native-quick-crypto

react-native-libraries.json
```diff
@@ -10990,9 +10990,12 @@
   },
   {
     "githubUrl": "https://github.com/margelo/react-native-quick-crypto",
+    "npmPkg": "react-native-quick-crypto",
     "examples": ["https://github.com/margelo/react-native-quick-crypto/tree/main/example"],
     "ios": true,
-    "android": true
+    "android": true,
+    "newArchitecture": true,
+    "newArchitectureNote": "use v1.0.0 or higher for new architecture support"
```
Simek, line 10998:
```suggestion
    "newArchitectureNote": "Use v1.0.0 or higher for new architecture support"
```

Casing nit.

boorad: done

boorad:
Makes sense.  We just reorg'd the repo for new arch, mono repo, etc.  Thanks for the help!

Simek (approved):
Thanks for the updates! 👍

PR #1277, opened 2024-09-17
Feat/update intercom in React Native Librairies

delphinebugner:
# 📝 Why & how
[Intercom](https://www.intercom.com) is a platform for support in mobile. This PR is to add their official bridge to the react native directory - it's maintained and working, I use it in my app; they have developed the correct Expo config plugin.  

As of New Architecture, I don't think they support it yet - I have'nt tried it, but I don't see in their release notes from the last past year anything related to supporting it. I intentionally let no boolean "New Architecture" to have the `untested` status displayed.  

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -1737,6 +1737,11 @@
     "android": true
   },
   {
+    "githubUrl": "https://github.com/tinycreative/react-native-intercom",
+    "ios": true,
+    "android": true
+  },
+  {
     "githubUrl": "https://github.com/joshswan/react-native-globalize",
     "ios": true,
     "android": true,
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -11899,5 +11899,26 @@
     "expoGo": true,
     "macos": true,
     "newArchitecture": true
+  }, 
+  {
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -1736,11 +1736,6 @@
     "githubUrl": "https://github.com/cesardeazevedo/react-native-bottom-sheet-behavior",
     "android": true
   },
-  {
-    "githubUrl": "https://github.com/tinycreative/react-native-intercom",
-    "ios": true,
-    "android": true
-  },
```
Simek, line 1743: If the package exists in npm registry we usually mark it as deprecated, instead of removing it, since it can be a good source of information for people to migrate off. The information for the directory API is also used by different tools, like `expo-doctor` to warn about and aid the migration.

We also support package suggestion, so in this case it would be:
```json
  {
    "githubUrl": "https://github.com/tinycreative/react-native-intercom",
    "ios": true,
    "android": true,
    "unmaintained": "@intercom/intercom-react-native"
  },
```

delphinebugner: Great insight thanks, I made the update 🙂

Simek (approved):
Thanks for the updates, LGTM! 🚀

PR #1287, opened 2024-09-21
update entry for `react-native-ios-context-menu`

dominicstop:
# 📝 Why & how

hello, and thank you for the hard work 😊
<br>
i would like to update the entry for [react-native-ios-context-menu](https://github.com/dominicstop/react-native-ios-context-menu)  to indicate that it now supports fabric/new arch via the `newArchitecture` flag

# ✅ Checklist


Simek (approved):
Thanks for the update! 👍

PR #1289, opened 2024-09-22
chore: update status of react-native-camera-kit

siddarthkay:
# 📝 Why & how
As reported here -> https://github.com/reactwg/react-native-new-architecture/discussions/167#discussioncomment-9674242 

`react-native-camera-kit` does not support new architecture yet.
I have updated this info in the directory.

# ✅ Checklist


Simek (approved):
Thanks for the update! 🚀

PR #1293, opened 2024-09-26
Update react-native-cloud-storage

mfkrause:
# 📝 Why & how
Adds the `"newArchitecture": true` flag to [react-native-cloud-storage](https://github.com/Kuatsu/react-native-cloud-storage). I've just tested the library on the New Architecture and, [after fixing an issue](https://github.com/Kuatsu/react-native-cloud-storage/commit/364baee1dd74a7ff8397295cd0cf8ef37e7a672e), confirmed it's now fully compatible through the Interop Layer in both bridged & bridgeless mode.

# ✅ Checklist


Simek (approved):
Thanks for the reference and update! 👍

PR #1328, opened 2024-10-19
add `navigation` to directory

gerzonc:
# 📝 Why & how

This PR adds `navigation` router to the directory

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -12217,7 +12217,8 @@
     ],
     "ios": true,
     "android": true,
-    "web": true
+    "web": true,
+    "newArchitecture": true
   },
   {
     "githubUrl": "https://github.com/grahammendick/navigation/tree/master/build/npm/navigation-react-native",
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -12202,5 +12202,16 @@
     "examples": ["https://github.com/nativeflowteam/nativeflowcss/tree/main/example"],
     "ios": true,
     "android": true
+  },
+  {
+    "githubUrl": "https://github.com/grahammendick/navigation",
+    "npmPkg": "https://www.npmjs.com/package/navigation",
+    "examples": [
+      "https://github.com/grahammendick/navigation/tree/master/NavigationReactNative/sample"
+    ],
+    "ios": true,
+    "android": true,
+    "web": true,
+    "newArchitecture": true
   }
 ]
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -12202,5 +12202,16 @@
     "examples": ["https://github.com/nativeflowteam/nativeflowcss/tree/main/example"],
     "ios": true,
     "android": true
+  },
+  {
+    "githubUrl": "https://github.com/grahammendick/navigation",
+    "npmPkg": "https://www.npmjs.com/package/navigation",
```
Simek, line 12208: Hello @gerzonc, thanks for adding the new entry!

There are two thing that need to corrected, before we can ship that addition - `npkPkg` field should match the package name, and the `githubUrl`, in monorepo, need to point directly to the directory when a `package.json` file for a given package resides.

```suggestion
    "githubUrl": "https://github.com/grahammendick/navigation/blob/master/build/npm/navigation",
```

Btw. The repository structure is quite unusual, those there might be some other minor problems, but we will try to address them! It might be a good idea to also add other packages from monorepo, like `navigation-react` and `navigation-react-native`, if they are required when using `navigation` lib in React Native apps.

gerzonc: Aah, I see, interesting, I'll take a look into fixing those issues, thanks!

react-native-libraries.json
```diff
@@ -12202,5 +12202,30 @@
...
+  },
+  {
+    "githubUrl": "https://github.com/grahammendick/navigation/tree/master/build/npm/navigation-react",
+    "examples": [
+      "https://github.com/grahammendick/navigation/tree/master/NavigationReact/sample"
+    ],
+    "ios": true,
+    "android": true,
+    "web": true
+  },
```
gerzonc, line 12221: should i add `newArchitecture` even for the dependencies like this one? `navigation-react-native` has all the stuff related to native stuff, not sure if to add new arch property to `navigation-react` as well, all it includes is just react code that works universally

Simek: Thanks for the updates! 👍 

> should i add newArchitecture even for the dependencies like this one?

Yes, please include that tag, it would just confirm the fact, and would boost the library score a bit in the migration period.

gerzonc: Done! Please take a look and please let me know if there is something else I might need to modify

Simek (approved):
Everything seems to be fine, thanks for the additions! Let's ship them! 🚀

PR #1335, opened 2024-10-25
Add react-native-multiple-modals

paufau:
# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -12306,5 +12306,15 @@
     "ios": true,
     "android": true,
     "expoGo": true
+  },
+  {
+    "githubUrl": "https://github.com/paufau/react-native-multiple-modals",
+    "npmPkg": "react-native-multiple-modals",
+    "examples": [
+      "https://github.com/paufau/react-native-multiple-modals?tab=readme-ov-file#usage",
+    ],
+    "ios": true,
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -12306,5 +12306,15 @@
     "ios": true,
     "android": true,
     "expoGo": true
+  },
+  {
+    "githubUrl": "https://github.com/paufau/react-native-multiple-modals",
+    "npmPkg": "react-native-multiple-modals",
+    "examples": [
+      "https://github.com/paufau/react-native-multiple-modals?tab=readme-ov-file#usage"
+    ],
```
Simek, line 12315: Hey @paufau, thanks for adding a new entry! 👍 

We can skip the `npmPkg` name since it matches the repository name (it's field mainly for scoped packages mainly), and the example link to project readme can be removed since package name link would lead to the repository anyway. Usually we expect example project or Snack in that field:
* https://github.com/react-native-community/directory?tab=readme-ov-file#examples

```suggestion
```

paufau: Thanks for your feedback! Updated PR to accommodate requested changes.

Simek (approved):
Thank you for the quick update! 🚀

PR #1352, opened 2024-11-08
Invertase package update roundup

mikehardy:
# 📝 Why & how

react-native-firebase:
- supports tvOS now (for most modules, where underlying SDK allows)
- new app-check and app-distribution packages

@invertase/react-native-apple-authentication
- qualified on new architecture
- default branch is main

react-native-google-mobile-ads
- supports new arch (and is mostly turbomodules now...)

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -2083,11 +2083,28 @@
       "https://github.com/invertase/react-native-google-mobile-ads/tree/main/RNGoogleMobileAdsExample"
     ],
     "ios": true,
-    "android": true
+    "android": true,
+    "newArchitecture": true
   },
   {
     "githubUrl": "https://github.com/invertase/react-native-firebase/tree/main/packages/analytics",
     "npmPkg": "@react-native-firebase/analytics",
+    "ios": true,
+    "tvos": true,
+    "android": true,
+    "newArchitecture": true
+  },
+  {
+    "githubUrl": "https://github.com/invertase/react-native-firebase/tree/main/packages/app-check",
+    "npmPkg": "@react-native-firebase/app-check",
+    "ios": true,
+    "tvos": true,
+    "android": true,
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -2083,40 +2083,61 @@
...
+    "npmPkg": "@react-native-firebase/app-check",
+    "ios": true,
+    "tvos": true,
+    "android": true,
+    "newArchitecture": true
+  },
+  {
+    "githubUrl": "https://github.com/invertase/react-native-firebase/tree/main/packages/app-distribution",
+    "npmPkg": "@react-native-firebase/app-distribution",
+    "ios": true,
```
Simek, line 2108: Hey @mikehardy, thanks for the updates! 👍 

Let's move new libraries at the end of file:
* https://github.com/react-native-community/directory?tab=readme-ov-file#how-do-i-add-a-library

mikehardy: May I please please request that they go here, since they are a monorepo of packages and I added them up with their siblings in alphabetical order? I see that the order is for "recently added" but having these packages appear as "fresh" (hey! recently added! type information for users) honestly has no value here in this case for us if that's the only reason.

If there is no way to bend on this then...okay

Simek: The new status means new to the directory, not fresh packages. Moving them down could help with discoverability, but it's fine if you prefer having them grouped.

Simek (approved):

react-native-libraries.json
mikehardy: Thanks! You inferred correctly in case my statement was ambiguous - there was no value in discoverability really, if people are using firebase they either know these exist already or they will just know they need them when they need them and they'll expect them to be there and they will be :-). Firebase and all its little services are like that

PR #1359, opened 2024-11-12
Mark lack of support for new arch on rive

tslater:
# 📝 Why & how
Mark lack of support for new arch on rive

# ✅ Checklist

Here is the clear evidence of lack of support:
https://github.com/rive-app/rive-react-native/issues/190


Simek (approved):
Thanks for the context and update! 👍

PR #1360, opened 2024-11-12
react-native-zoom-reanimated: add npm

kesha-antonov:
Hi

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -9796,6 +9796,7 @@
   },
   {
     "githubUrl": "https://github.com/kesha-antonov/react-native-zoom-reanimated",
+    "npmPkg": "react-native-zoom-reanimated",
     "examples": ["https://github.com/kesha-antonov/react-native-zoom-reanimated/tree/main/example"],
     "ios": true,
     "android": true,


Simek (commented):
Hey @kesha-antonov, this should not be need unless it's a scoped package or repository has been converted to the monorepo. Can you explain the context of the proposed change?

If the reason is lack of download stats, it can be attributed to recent npm API degradation, which also affects fresh packages validation.

kesha-antonov:
> Hey @kesha-antonov, this should not be need unless it's a scoped package or repository has been converted to the monorepo. Can you explain the context of the proposed change?
> 
> If the reason is lack of download stats, it can be attributed to recent npm API degradation, which also affects fresh packages validation.

I don't see downloads stats here: https://reactnative.directory/?search=react-native-zoom-reanimated

Screenshot 2024-11-12 at 18 03 56

If we have link to npm then we see downlods stats

Screenshot 2024-11-12 at 18 04 10

Simek:
It's not due to lack of `npmPkg`, it's due to some recent degradation/changes on the npm API side. 

It looks like non-scoped packages download data started to return no data starting from few weeks ago, and we use different method of data fetching based on the package type since bulk queries does not support scoped packages:
* https://github.com/npm/registry/blob/main/docs/download-counts.md#bulk-queries

As said in the comment above, this also affect some of the new entries waiting for being accepted, will investigate the root of this soon.

Simek:
Screenshot 2024-11-12 at 20 25 15

@kesha-antonov The npm download data is back for most of the entries. 

I will also follow up on fetch changes with a fallback to previously stored data, if download stats are present in there.

kesha-antonov:
> Screenshot 2024-11-12 at 20 25 15
> 
> 
> 
> @kesha-antonov The npm download data is back for most of the entries. 
> 
> 
> 
> I will also follow up on fetch changes with a fallback to previously stored data, if download stats are present in there.

Thanks!

PR #1379, opened 2024-11-20
Fix: Remove macOS support from `react-native-nitro-sqlite`

chrispader:
# 📝 Why & how

During my previous PR which adds `react-native-nitro-sqlite` i added `"macos": true` to NitroSQLite listing. This was added by accident, as of now the library doesn't have official macOS support.

# ✅ Checklist


Simek (approved):
Thank you for the follow-up! 👌

PR #1395, opened 2024-11-24
feat: add react compiler runtime

KaBankz:
# 📝 Why & how

- Add react-compiler-runtime

> [!NOTE]
> I haven't tested the react compiler on any targets besides web and dev clients on ios and android. As such, I've left support for all other targets as false.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -12892,5 +12892,12 @@
     "examples": ["https://github.com/ttlock/react-native-ttlock/tree/master/example"],
     "ios": true,
     "android": true
+  },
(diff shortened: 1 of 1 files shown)


Simek (approved):
Thanks for the addition! 👍

KaBankz:
I just realized that I didn't add `"npmPkg": "react-compiler-runtime"`

KaBankz:
@Simek The pre commit hooks are not letting me set the `npmPkg` key.

I thought it was necessary in cases like this package?

PR #1397, opened 2024-11-25
feat: Add Nitro Modules (and nitrogen)

mrousavy:
# 📝 Why & how

I added Nitro Modules and nitrogen to this list. 

There are already two libraries using both Nitro Modules and Nitrogen in the directory.

# ✅ Checklist


mrousavy:
btw @Simek how does one library get the "Recommended" library badge? I wonder if react-native-mmkv or react-native-vision-camera are eligible for that... 👀

Simek (commented):
Hey @mrousavy, thanks for the new additions! 👍 

Can you move the newly added libraries at the end of file? We use that for one of sort orders, it also helps a bit with discoverability:
* https://github.com/react-native-community/directory?tab=readme-ov-file#how-do-i-add-a-library

---

When it come to the "Recommended"  flag, we have a definition of the criterion on the scoring page:
* https://reactnative.directory/scoring

Screenshot 2024-11-25 at 19 49 24

But to shed a bit more light into the topic, we did not re-evaluated internally for a longer while to which libraries we should add, or from which to remove the status. Because with time, we saw more confusion on the directory users side, than benefits which this flags could potentially give.

We have some plans in mind related that one hand-picked status, since a lot things have changed since we initially added it. Stay tuned for some follow ups in the future! 🙂

baronha:
Marc. Instead of adding Nitro Modules, add nitro modules next to other indexes such as iOS, Android, Expo Go. 
Let's goooo

react-native-libraries.json
```diff
@@ -13123,6 +13123,38 @@
     "ios": true,
     "android": true,
     "expoGo": true,
+  },
+    "githubUrl": "https://github.com/mrousavy/nitro/tree/main/packages/react-native-nitro-modules",
```
Simek, line 13127:
```suggestion
  },
  {
    "githubUrl": "https://github.com/mrousavy/nitro/tree/main/packages/react-native-nitro-modules",
```

The opening bracket is missing, which throws syntax error in CI.

mrousavy: sorry my bad did the merge on my phone

react-native-libraries.json
```diff
@@ -13123,6 +13123,39 @@
     "ios": true,
     "android": true,
     "expoGo": true,
+  },
```
Simek, line 13126:
```suggestion
    "newArchitecture": true
  },
```

Looks like you have removed new architecture flag from previous entry, and now trailing comma on Expo Go line throws.

Simek (approved):
Validated the changes by running the checks locally, thanks for additions and updates! 👍

PR #1401, opened 2024-11-28
Add i18n-js (#1401)

alexisgahon:


react-native-libraries.json
```diff
@@ -12935,5 +12935,12 @@
     "ios": true,
     "android": true,
     "newArchitecture": true
+  },
+  {
+    "githubUrl": "https://github.com/fnando/i18n",
+    "ios": true,
```
Simek, line 12941: Hello @alexisgahon, thanks for adding new entry to the directory! 👍 

Looks like this library name differs from the repository name, in those cases we can specify "npmPkg" manually:

```suggestion
    "githubUrl": "https://github.com/fnando/i18n",
    "npmPkg": "i18n-js",
    "ios": true,
```

Simek (approved):
Thanks for the update, LGTM! 🚀

Issue #1376, opened 2024-11-19
Wrong "unmaintained" label


BrodaNoel:
I was checking that my library, https://reactnative.directory/?search=duix, which is maintained, is being shown as `unmaintained`, even considering that it is not defined as unmaintained in the https://github.com/react-native-community/directory/blob/main/react-native-libraries.json file.

```
  {
    "githubUrl": "https://github.com/BrodaNoel/duix",
    "ios": true,
    "android": true,
    "web": true,
    "expoGo": true,
    "newArchitecture": true
  },
```

Or should I add the `unmaintained: false` attribute?

Simek:
Hey @BrodaNoel, 

We mark libraries automatically as "unmaintained" if there was no code related activity in the project repository in the last two years:
* https://github.com/react-native-community/directory/blob/0b0968db7ff4927ba400532c68cdfce642eaf38d/scripts/build-and-score-data.js#L65

The latest commit you have pushed few hours ago fixed the categorization already. The "unmaintained" label to that point was correct according to the logic, since the second to last commit was made in April 2022. 

![Screenshot 2024-11-19 at 11 06 59](https://github.com/user-attachments/assets/4f24e0e5-aec3-47f4-bda8-db5da07f86b7)

PR #1406, opened 2024-12-01
Change "Updated" metadata field to be a link to commit history

esthor:
# 📝 Why & how

I noticed the "Updated" metadata field didn't have a link. It is referencing [`pushedAt`](https://arc.net/l/quote/ympnxekk) from [Github's API for Repository](https://docs.github.com/en/graphql/reference/objects#repository) which:
>  Identifies the date and time when the repository was last pushed to.

I thought then it makes sense, as someone who wants to go a little deeper into "Updated" / see the receipts, that the most logical link would be to the commits. Plus, it saves a couple clicks when you are evaluating a library and go to see how frequently it's updated.

| Before | After |
|--------|--------|
| image | image |

#### Other considerations:

A bigger change would be to instead calculate "Updated" based on the last push of **the npm package**, since changes to a git repo aren't necessarily reflected in when the actual package (e.g. from a `npm install package-123`) was last updated. However, this would change what this metadata value means significantly and would be breaking. It would also require changing the Scoring algorithm to look at npm package dates. And, in the cases where you _really_ care about how quickly a package is updated (e.g. everything's on fire until a fix it ready), you could always point to the source of truth (github repo code) instead of the npm package, meaning you care mostly about how quick to respond and fix the maintainers are. I'm new to this repo, so this has probably already been thought about, so I'm just sharing my initial thought process because I went back-and-forth on it and I'm sure others might too. 🤷 

# ✅ Checklist

### DIFF
--- a/components/Library/MetaData.tsx
+++ b/components/Library/MetaData.tsx
@@ -44,9 +44,9 @@ function generateData({ github, score, npm, npmPkg }: LibraryType, isDark: boole
       id: 'calendar',
       icon: <Calendar fill={iconColor} />,
       content: (
-        <Caption style={{ fontWeight: 300 }}>
+        <A href={`${github.urls.repo}/commits`} style={styles.link}>
           Updated {getTimeSinceToday(github.stats.pushedAt)}
-        </Caption>
+        </A>
       ),
     },
     npm.downloads


Simek:
Hey @esthor, thanks for proposing that change! I think that linking the date is a good idea. 👍 

Also thank you for the feedback! 🙏 

When it comes to your consideration - we have made conscious decision what is the source of "updated" stat based on testing different approaches a while ago. Using npm package latests release date have it pros and cons, likewise as current approach, but we wanted to give the library authors and maintainers a bit more leniency, and end-users a way more dynamic list of recently updated libraries.

Also because we mark libraries as "unmaintained" partially based on that date, we wanted to avoid having false-positives, since this state affects library score and presentation, and still there are cases when we have such cases, i.e.:
* https://github.com/react-native-community/directory/issues/1376

In the future, we plan to expand the library page, and include more npm releases/version information, including the dates, which might also help a bit with evaluating the exact package maintenance state. It would probably also become a part of scoring algorithm at that point, so the future goal would be to have/store/use both of the dates, not replacing one with other.

Simek (approved):

PR #1411, opened 2024-12-03
flag react-native-maps as new arch incompatible

fobos531:
# 📝 Why & how

As per [this comment](https://github.com/react-native-maps/react-native-maps/issues/5206#issuecomment-2513067290) of react-native-maps' maintainer, let's indicate that the library is new arch incompatible. While it CAN run with the interop layer, it has a plethora of issues that need to be resolved by properly migrating the library and its views to Fabric & new arch. Highlighting this library as new arch compatible (even more so that its highlighted as "recommended library") will only encourage usage of the library with new arch, putting even more strain on the `react-native-maps` maintainers while they are in the middle of migrating - just check the issue board on that repo, it's flooded with new arch issues. 

With that said, it might be worth to indicate incompatibility with new arch in react native directory, and only reenable it once the compatibility has been widely confirmed by the community.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -578,7 +578,7 @@
       "https://snack.expo.dev/H1zOFxnN-"
     ],
     "goldstar": true,
-    "newArchitecture": true
+    "newArchitecture": false
   },
   {
     "githubUrl": "https://github.com/bramus/react-native-maps-directions",


Simek (approved):
Hey @fobos53, thank you for providing a context on this update, LGTM! 👍

PR #1466, opened 2025-01-08
Add `redux`

aryaemami59:
# 📝 Why & how

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -13355,5 +13355,14 @@
     "web": true,
     "expo": true,
     "newArchitecture": true
+  },
+  {
+    "githubUrl": "https://github.com/reduxjs/redux",
+    "examples": ["https://github.com/reduxjs/redux/tree/master/examples"],
+    "ios": true,
+    "android": true,
+    "web": true,
+    "expo": true,
+    "newArchitecture": true
   }
 ]


react-native-libraries.json
```diff
@@ -13330,5 +13330,17 @@
     "ios": true,
     "android": true,
     "newArchitecture": false
+  },
+  {
+    "githubUrl": "https://github.com/reduxjs/redux",
+    "examples": [
+      "https://github.com/reduxjs/redux-toolkit/tree/HEAD/examples/publish-ci/react-native",
+      "https://github.com/reduxjs/redux-toolkit/tree/HEAD/examples/publish-ci/expo"
+    ],
```
Simek, line 13339: Redux repo has its own examples which might be better to reference than examples from 3rd party repo:
* https://github.com/reduxjs/redux/tree/master/examples

The code in `redux-toolkit` also does not directly specify Redux as dependencies.

aryaemami59: I'm a little confused because `redux` is listed as a [**direct dependency**](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/package.json#L127) of `@reduxjs/toolkit`. Also the examples listed are not from a 3rd party repo, they're from the official `redux-toolkit` monorepo.

aryaemami59: I modified the `examples` array.

Simek (approved):

PR #1484, opened 2025-01-21
add web support for react-native-video

xilin:
# 📝 Why & how

`react-native-video` added web support in https://github.com/TheWidlarzGroup/react-native-video/pull/3958, so I would like to add `web: true` here

# ✅ Checklist


Simek (approved):
Thanks for the update, and providing a context for the change! 👍

PR #1498, opened 2025-02-02
Add xstate, @xstate/store and @xstate/react

wodin:
# 📝 Why & how
This adds xstate and the related @xstate/store and @xstate/react. There are other related libraries, but I haven't used those.

# ✅ Checklist


Simek (approved):
Thanks for adding new libraries @wodin! 👌

PR #1499, opened 2025-02-03
new libary - expo-key-event

tlow92:


react-native-libraries.json
```diff
@@ -13680,5 +13680,16 @@
     "android": true,
     "web": true,
     "expoGo": true
+  },
+  {
+    "githubUrl": "https://github.com/tlow92/expo-key-event",
+    "npmPkg": "https://www.npmjs.com/package/expo-key-event",
```
Simek, line 13686: Hey @tlow92, thanks for adding new entry to the directory!

The `npmPkg` package field expect passing the name of published npm package in the registry. In case of your entry it is also redundant, since repository name matches package name:
* https://github.com/react-native-community/directory?tab=readme-ov-file#npmpkg

```suggestion
```

tlow92: You are right, I missed that. Thank you for the feedback 🙏

Simek (approved):
Thanks for the update, LGTM! 👍

PR #1500, opened 2025-02-03
Add react-native-mqtt-file-transfer

benjaminkomen:
# 📝 Why & how

- Add new React Native module `react-native-mqtt-file-transfer`, which can be used to send files over MQTT directly from the native layer, greatly speeding up the file transfer

# ✅ Checklist

TODO: make github repository public

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -13665,5 +13665,13 @@
     "android": true,
     "web": true,
     "expoGo": true
+  },
+  {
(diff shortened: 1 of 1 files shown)


Simek:
Hey @benjaminkomen, unfortunately we do not support closed-source libraries, since a lot of metrics/metadata is used for categorizing and scoring. We also perform automatic code analysis, which is not possible with private repositories.

benjaminkomen:
Understood, I will try to make it public, that's why the PR is still a draft

PR #1503, opened 2025-02-05
fix: refactor filters for easier maintenance

douglowder:
# 📝 Why & how

- Refactor the code in `components/Filters` to avoid hardcoding filter params in `index.tsx`
- More clearly distinguish between platforms and other filters
- Clearly distinguish and group filters that require main search path

Screenshot 2025-02-04 at 8 16 50 PM

### DIFF
--- a/components/Filters/helpers.ts
+++ b/components/Filters/helpers.ts
@@ -4,10 +4,6 @@ export const FILTER_PLATFORMS = [
     title: 'Android',
   },
   {
-    param: 'expoGo',
-    title: 'Expo Go',
-  },
-  {
     param: 'ios',
     title: 'iOS',
   },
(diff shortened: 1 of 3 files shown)


components/Filters/helpers.ts
```diff
@@ -32,3 +28,45 @@ export const FILTER_PLATFORMS = [
...
+    param: 'expoGo',
+    title: 'Works with Expo Go',
+  },
+  {
+    param: 'hasTypes',
+    title: 'Has Typescript types',
+  },
+  {
+    param: 'newArchitecture',
+    title: 'Supports new architecture',
```
Simek, line 66:
```suggestion
    title: 'Supports New Architecture',
```

components/Filters/helpers.ts
```diff
@@ -32,3 +28,45 @@ export const FILTER_PLATFORMS = [
...
+    param: 'hasImage',
+    title: 'Has image preview',
+  },
+  {
+    param: 'expoGo',
+    title: 'Works with Expo Go',
+  },
+  {
+    param: 'hasTypes',
+    title: 'Has Typescript types',
```
Simek, line 62:
```suggestion
    title: 'Has TypeScript types',
```

components/Filters/helpers.ts
```diff
@@ -32,3 +28,45 @@ export const FILTER_PLATFORMS = [
...
+    param: 'isPopular',
+    title: 'Popular',
+  },
+  {
+    param: 'isRecommended',
+    title: 'Recommended',
+  },
+];
+
+export const FILTER_STATUS = [
```
Simek, line 47: Since we already alter the code, can we also reorder those filters?

Suggested order:
* new arch
* example
* images
* types
* expo go
* updated

douglowder: Done

Simek (approved):
Thanks for those improvements! 👍 

Will follow up with statuses display tweaks, so we no longer list Expo Go along side supported platforms.

PR #1510, opened 2025-02-12
Add react-native-segmented-control-2

seanblonien:
# 📝 Why & how

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -13673,7 +13673,9 @@
       "https://raw.githubusercontent.com/EdgarJMesquita/expo-totp/refs/heads/main/docs/assets/demo.gif"
     ],
     "ios": true,
-    "android": true
+    "android": true,
+    "expoGo": false,
+    "web": false
   },
   {
     "githubUrl": "https://github.com/sellpy/react-native-scroll-anchor",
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -13695,5 +13695,29 @@
...
+    ],
+    "images": [
+      "https://github.com/WrathChaos/react-native-segmented-control-2/blob/master/assets/Screenshots/react-native-segmented-control-2.gif"
+    ],
+    "alternatives": [
+      "@react-native-segmented-control/segmented-control",
+      "rn-segmented-tab-controls",
+      "react-native-paper",
+      "react-native-segmented-control-tabs"
+    ],
```
Simek, line 13712: We only use `alternatives` for deprecated or unmaintained libraries, for which author made a direct recommendation:

react-native-libraries.json
```diff
@@ -13695,5 +13695,29 @@
...
+    "ios": true,
+    "android": true,
+    "web": true,
+    "windows": true,
+    "macos": true,
+    "tvos": true,
+    "visionos": true,
```
Simek, line 13719: Adding out-of-tree platforms support requires a dedicated example, or reference to an OSS app using it on that platform:
* https://github.com/react-native-community/directory?tab=readme-ov-file#%EF%B8%8F-out-of-tree-platforms

I can only see the regular example for build-in native platforms in the repository code. To do not block the addition, we can update the entry later in time, when we have some references for out-of-tree usage.

Simek (commented):
Hey @seanblonien, thanks for adding new entry to the directory! 👍Left few suggestions below.

react-native-libraries.json
seanblonien: Ahh i see, noted and removed

react-native-libraries.json
seanblonien: ahh ok, i will just leave ios, android, and web then

Simek (approved):
Thanks for the updates, let's ship this! 👍

PR #1512, opened 2025-02-14
 Add Fire OS compatibility

brentvatne:
Rebased #1438 and following up here


brentvatne:
fixed by original author on original pr

PR #1532, opened 2025-03-05
Mark legend list as working in new arch and expo go

kadikraman:
# 📝 Why & how

Double checked with Jay and [Legend List](https://github.com/LegendApp/legend-list) is very much tested in new arch.

It's a JS-only library and also works in Expo Go so I added that as well.

# ✅ Checklist


Simek (approved):
Thanks! 👍

PR #1535, opened 2025-03-06
Fix new architecture detection

nympheastudio:
This pull request introduces improvements to the detection of libraries supporting the new architecture in React Native. Specifically, it addresses issues where some libraries were incorrectly marked as supporting the new architecture, even though they do not.

# 📝 Why & how

This update ensures that libraries which are outdated, unmaintained, or incompatible with React Native's new architecture are properly marked as unsupported. It also provides more accurate detection of new architecture support based on library metadata, React Native version, and maintenance status.

# 🔢  Example  of bug  : 
On https://reactnative.directory/?search=maps, react-native-map-clustering is incorrectly tagged with "New Architecture," even though it doesn't actually support it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,15187 @@
+{
+  "name": "directory",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "dependencies": {
+        "@expo/html-elements": "^0.11.0",
+        "@expo/match-media": "^0.4.0",
+        "@jlengstorf/get-share-image": "^1.0.2",
+        "@radix-ui/react-hover-card": "^1.1.2",
+        "@radix-ui/react-tooltip": "^1.1.3",
+        "@react-native-async-storage/async-storage": "^2.1.0",
+        "@react-native-picker/picker": "^2.10.2",
+        "@sentry/react": "^8.46.0",
+        "@vercel/blob": "^0.27.0",
+        "expo": "^52.0.23",
+        "expo-font": "~13.0.2",
+        "lodash": "^4.17.21",
+        "next": "^15.1.3",
+        "node-emoji": "^2.2.0",
+        "react": "^18.3.1",
+        "react-content-loader": "^7.0.2",
+        "react-dom": "^18.3.1",
+        "react-easy-linkify": "^1.0.8",
+        "react-native": "^0.76.5",
+        "react-native-safe-area-context": "^4.14.1",
+        "react-native-svg": "15.4.0",
+        "react-native-web": "^0.19.13",
+        "react-native-web-hooks": "^3.0.2",
+        "use-debounce": "^10.0.4"
+      },
+      "devDependencies": {
+        "@expo/next-adapter": "^6.0.0",
+        "@next/bundle-analyzer": "^15.1.3",
+        "@types/lodash": "^4.17.15",
+        "@types/node": "^22.10.2",
+        "@types/react": "^18.3.12",
+        "ajv-cli": "^5.0.0",
+        "browserslist": "^4.24.3",
+        "cheerio": "^1.0.0",
+        "cross-fetch": "^4.1.0",
+        "dotenv": "^16.4.7",
+        "eslint": "^8.57.0",
+        "eslint-config-next": "^15.1.3",
+        "eslint-config-universe": "14.0.0",
+        "lint-staged": "^15.2.11",
+        "next-compose-plugins": "^2.2.1",
+        "next-fonts": "^1.5.1",
+        "next-images": "^1.8.5",
+        "prettier": "^3.4.2",
+        "simple-git-hooks": "^2.11.1",
+        "typescript": "5.6.3",
+        "user-agent-data-types": "^0.4.2"
+      }
+    },
+    "node_modules/@0no-co/graphql.web": {
+      "version": "1.1.2",
+      "resolved": "https://registry.npmjs.org/@0no-co/graphql.web/-/graphql.web-1.1.2.tgz",
+      "integrity": "sha512-N2NGsU5FLBhT8NZ+3l2YrzZSHITjNXNuDhC4iDiikv0IujaJ0Xc6xIxQZ/Ek3Cb+rgPjnLHYyJm11tInuJn+cw==",
+      "peerDependencies": {
+        "graphql": "^14.0.0 || ^15.0.0 || ^16.0.0"
+      },
+      "peerDependenciesMeta": {
+        "graphql": {
+          "optional": true
+        }
+      }
+    },
+    "node_modules/@ampproject/remapping": {
+      "version": "2.3.0",
+      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
+      "integrity": "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==",
+      "dependencies": {
+        "@jridgewell/gen-mapping": "^0.3.5",
+        "@jridgewell/trace-mapping": "^0.3.24"
+      },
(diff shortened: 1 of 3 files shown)


package-lock.json
Simek, line 1: Remove from chanegest.

util/newArchStatus.ts
```diff
@@ -6,22 +6,52 @@ export enum NewArchSupportStatus {
...
 
-export function getNewArchSupportStatus({ newArchitecture, github, expoGo }: Library) {
+//new function  to fix new architecture support detection
+export function getNewArchSupportStatus({
+  newArchitecture,
+  github,
+  expoGo,
+  lastPublish,
+  maintained,
+  reactNativeVersion,
```
Simek, line 16: These does not exist, you can see what information is available by looking at `Library` type.

Screenshot 2025-03-10 at 11 12 51

util/newArchStatus.ts
```diff
@@ -6,22 +6,52 @@ export enum NewArchSupportStatus {
...
+}: Library) {
   // Assume untested unless indicated otherwise through one of the following tests
   let flag = undefined;
 
+  // If the library is not maintained, it likely doesn't support the New Architecture
+  if (maintained === false) {
+    return NewArchSupportStatus.Unsupported;
+  }
```
Simek, line 24: Unfortunately, this is an incorrect assumption. Unmaintained libraries can support New Architecture in multiple cases, like:
* deprecation has happen after New Architecture introduction
* package is JS only (does not include native code)
* package has been marked as unmaintend due to name change

util/newArchStatus.ts
```diff
@@ -6,22 +6,52 @@ export enum NewArchSupportStatus {
...
+
+  // Check if the last publish date is before the introduction of the New Architecture
+  // The New Architecture was introduced around React Native 0.68 (April 2022)
+  const newArchIntroDate = new Date('2022-04-01');
+  if (lastPublish && new Date(lastPublish) < newArchIntroDate) {
+    return NewArchSupportStatus.Unsupported;
+  }
```
Simek, line 31: This is not true for package not including any native code (JS only).

util/newArchStatus.ts
```diff
@@ -6,22 +6,52 @@ export enum NewArchSupportStatus {
...
+
+  // Check if the React Native version in the library's package.json is older than 0.68
+  // If it is, the library probably doesn't support the New Architecture
+  if (reactNativeVersion && isVersionOlderThan(reactNativeVersion, '0.68.0')) {
+    return NewArchSupportStatus.Unsupported;
+  }
```
Simek, line 37: We do not have the React Native version information at this time. We were experimenting with this idea, we can extract the information for `package.json` file, but the problem is with the data itself, which package developers provide, and we have concluded that we cannot rely globally on that information.

What we have found out - often React Native is not specified as regular, peer or dev dependency, often when specified semver is ster to `*` which does not give us any viable information, there are also cases when this information is set once, and never updated, so we have entries like `>0.49`.

Simek (commented):
Hello @nympheastudio, thanks for bringing up an issue and attempting to address it! 👍 

In would be nice if you can provide multiple examples of incorrect packages marking, which would help us to paint more broad picture of New Architecture support issues you are trying to address.

If this more a case of one, or up to few libraries which are mismatch, we probably should just alter their entries instead, and set `newArchitecture: false` if needed.

After diving into diffs, looks like you did not run the code locally, since updated helper uses non-existing fields from `Library` after the updates. Sorry that we do not have workflows setup on the CI for basic code checks which will help you to caught that, will add that soon!

Simek:
With https://github.com/react-native-community/directory/pull/1539 shipped, when you rebase your PR you should be able to see the missing data issues on the CI.

nympheastudio:
Hello @Simek,

Thanks for the feedback!

Regarding the libraries I’ve found to be either not working or unstable with the new architecture, here are a few examples:

- react-native-map-clustering
- react-native-web-maps
- react-native-open-maps
- react-native-maps-directions
- expo-maps-polygon-editor
- maplibre-react-native
- react-native-buttonex
- react-native-responsive-image
- rn-displayable
- ex-navigation

if you visit https://reactnative.directory/?order=updated&direction=ascending&search=expo&newArchitecture=true you’ll see that almost every result there is not working on the new architecture. This should provide a clearer picture of the widespread issues I’m trying to address.

Don't you think that the logic if (expoGo === true) might no longer be considered valid, especially after the recent changes in Expo (versions 51 and 52) and React Native?

Regarding the current solution, I realize that my initial approach could be improved. Thanks for pointing out the issue with the helper using non-existing fields after the updates. 
I’ll work on a better fix, test it properly locally, and ensure everything is aligned with the current code.


I’ll keep you updated with my progress!

have nice day Simek

Simek:
Hey @nympheastudio, thanks for the answer, and a bit more context! 👍 

I have made a quick test project with some of the libraries from the list you have provided, which does not require additional setup steps, and I can definitely tell that they are working with New Architecture enabled correctly. They are mostly non-native code libraries, so this is expected.

Here is the source code:
* https://github.com/Simek/new-arch-test
* New Architecutre is enabled:
  * https://github.com/Simek/new-arch-test/blob/5aa4ff9905318ddd8408ec26ed3accd4383063ce/app.json#L10
* The list of dependencies I have tested:
  * https://github.com/Simek/new-arch-test/commit/5aa4ff9905318ddd8408ec26ed3accd4383063ce#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

Preview video from the simulator:

https://github.com/user-attachments/assets/8a6428a3-6160-4d8c-aec6-8442a3e9bc34

This research effort makes me think that there might something wrong with the app you trying to use those packages in, not with the packages itself.

> Don't you think that the logic if (expoGo === true) might no longer be considered valid, especially after the recent changes in Expo (versions 51 and 52) and React Native?

Nope, JavaScript only packages should work the same in older and the newest React Native versions. The New Architecture changes should not affect the compatibility of these packages. There might be a case or two overall where some other breaking changes (not related to New Architecture effort) have been made to React Native core, which resulted in libraries incompatibility, but this is a natural thing with versioning.

In general, if you are experiencing any problem with packages in ecosystem, the best approach is to work on minimal reproducible example, and then share that or report as an issue in the given package repository. You can read a bit more about it here:
* https://stackoverflow.com/help/minimal-reproducible-example

Only if we have confidence, or an example showing that the given package does not work with New Architecture, we should alter the state in directory. When it comes to the detection mechanism, I think it's in a quite good spot when it comes to automarking rules, but I you will be able to provide the more compelling arguments, with examples showing false-positive reports that are recurring, I'm happy to alter the logic we use.

nympheastudio:
Hey @Simek, thanks again for your response and for testing this out!

After testing your sample project, I noticed that Cluster is not working. The issue seems to come from react-native-maps, which, as mentioned in [this issue](https://github.com/react-native-maps/react-native-maps/issues/5161), does not yet support the New Architecture.

Also, I’d like to apologize for my previous doubts about react-native-responsive-image and react-native-buttonex—after further testing, they do indeed seem to support the New Architecture properly.

The main concern here is that a clustering library is marked as compatible with the New Architecture, even though it relies on react-native-maps, which is not yet fully compatible. While the package may load under Expo + new Arch and not crash, the fact that it cannot be used due to its dependency raises a question: wouldn’t it be misleading for readers of the directory to see it marked as compatible when, in practice, it doesn’t function properly?

To highlight this issue further, I’m going to test more packages listed in [[reactnative.directory](https://reactnative.directory/?order=updated&direction=ascending&search=expo&newArchitecture=true)](https://reactnative.directory/?order=updated&direction=ascending&search=expo&newArchitecture=true) to see if similar cases arise. I'll provide more detailed feedback soon. 

Thanks and have a good day.

nympheastudio:
i m still working on a solution, hope publish this or next week, thanks and have a nice day simek

Simek:
Hey @nympheastudio! 👋  Are you still interested in working on this PR?

PR #1543, opened 2025-03-12
Add macos support to legend-list

jmeistrich:
# 📝 Why & how
Adds macos support to legend-list.

I don't have a public example yet, but I've tested it extensively in macos and it works well. It should also work on every other out of tree platform since it's just JS, but I'll make an example app for every platform before I add those here.

# ✅ Checklist


Simek (approved):
Hey @jmeistrich, thanks for the update! 👍

PR #1555, opened 2025-03-24
Update the Navigation router to reference just the one navigation package

grahammendick:
# 📝 Why & how

[The Navigation router was added to the react native directory](https://github.com/react-native-community/directory/pull/1328) as 3 separate packages - navigation, navigation-react and navigation-react-native. But, after [chatting with @Simek on twitter](https://x.com/Simek/status/1849093053304201415) we decided it makes more sense to just add the navigation package and remove navigation-react and navigation-react-native. He suggested some changes to the [navigation npm README](https://www.npmjs.com/package/navigation) which I've also now done.

image

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -12336,27 +12336,14 @@
     "newArchitecture": true
   },
   {
-    "githubUrl": "https://github.com/grahammendick/navigation/tree/master/build/npm/navigation",
-    "ios": true,
-    "android": true,
-    "web": true,
-    "newArchitecture": true
-  },
-  {
-    "githubUrl": "https://github.com/grahammendick/navigation/tree/master/build/npm/navigation-react",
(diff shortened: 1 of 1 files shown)


Simek (approved):
Hey @grahammendick, thanks for the follow up updates! 👍 

Since we put more effort into the new architecture support compatibility in last months, and it become part of Expo validation process, the side-effect of removing packages for Directory will result in warnings during installation (using `expo install`) or when using `expo-doctor` by end users:
* https://docs.expo.dev/guides/new-architecture/#validate-your-dependencies-with-react-native-directory

The warnings can be silenced on the user-side, but there might be some community contributors which would like to add missing packages data to the directory, and we would need to refer them to this PR, to explain why those packages should not be in the directory.

If this is fine for you, I more than happy to ship the changes. 🙂

PR #1562, opened 2025-03-30
Add @config-plugins/react-native-blob-util, @config-plugins/react-native-pdf and crypto-js

SimranDev:
# 📝 Why & how
Libraries added:
- [@config-plugins/react-native-blob-util](https://github.com/expo/config-plugins/tree/main/packages/react-native-blob-util)
- [@config-plugins/react-native-pdf](https://github.com/expo/config-plugins/tree/main/packages/react-native-pdf)
- [crypto-js](https://github.com/brix/crypto-js)

# ✅ Checklist

- [x] Added libraries to **`react-native-libraries.json`**

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -14211,5 +14211,12 @@
     "githubUrl": "https://github.com/jeremybarbet/expo-testflight",
     "ios": true,
     "newArchitecture": true
+  },
+  {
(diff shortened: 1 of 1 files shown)


Simek (approved):
Thank you for adding new entries to the directory! 👍 

Side note: we might think about adding `configPlugin` field, so if one exists, especially 3rd party one, it's easier to link them with corresponding library.

PR #1563, opened 2025-03-31
fix(docs): calculate-score.js -> .ts broken link

gunnartorfis:
# 📝 Why & how

The links on the [scoring](https://reactnative.directory/scoring?utm_source=chatgpt.com) page are broken, they are linking to the old .js file. This change updates those links to be referencing the .ts file.


Simek (approved):
Nice catch, thanks for the fix! 👍

PR #1566, opened 2025-04-01
Update `rive-react-native`'s New Architecture Compatibility via Interop Layer

alextoudic:
# 📝 Why & how

`rive-react-native` is still labeled as incompatible with the new architecture, even though it now works thanks to the interop layer. Their example app is built with Expo and has the new architecture enabled: https://github.com/rive-app/rive-react-native/blob/8db519e65f5cfad91f7286c7c3c03334665b18c0/example/app.json#L10

# ✅ Checklist


Simek (approved):
Nice catch, thanks for the update! 👌

PR #1592, opened 2025-04-20
feat(directory): add better-auth to list

mezotv:
# 📝 Why & how
Adds both the better-auth packages to the index, just integrated better-auth into my app and it works seamingly accross devices but expo doctor doesn't know about it yet hence my PR!

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -14395,5 +14395,22 @@
     "ios": true,
     "android": true,
     "newArchitecture": true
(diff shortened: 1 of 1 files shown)


mezotv:
Totally missed that their main package.json has a totally different package name, should work now!

Simek (approved):
Hey @mezotv, thanks for adding new entries to the directory! 👍

PR #1602, opened 2025-04-26
Add tvos tag to many libraries that support Apple TV

douglowder:
# 📝 Why & how

The `tvos` tag is missing from many packages that support Apple TV. This PR adds those tags.


Simek (approved):
Thanks for the update Doug! 🙏

PR #1625, opened 2025-05-06
Added alternatives for visibility sensor and tree view libraries

JairajJangle:
# 📝 Why & how

Added alternatives for visibility sensor and tree view libraries. The added alternatives support new architecture and are well-maintained. 

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -14636,5 +14636,72 @@
     "android": true,
     "web": true,
     "expoGo": true
+  },
+  {
+    "githubUrl": "https://github.com/rudderlabs/rudder-sdk-react-native/tree/develop/libs/sdk",
+    "npmPkg": "@rudderstack/rudder-sdk-react-native",
+    "ios": true,
+    "android": true,
+    "tvos": true
+  },
+  {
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -12956,7 +12957,8 @@
       "https://raw.githubusercontent.com/SimformSolutionsPvtLtd/react-native-tree-selection/master/assets/Tree-Selection.gif"
     ],
     "ios": true,
-    "android": true
+    "android": true,
+    "alternatives": ["react-native-tree-multi-select"]
```
Simek, line 12961: Hey @JairajJangle, thanks for the updates! 👍 

Usually, we add alternatives only to libraries marked as no longer maintained. This one seems to still be out of the threshold:
* https://reactnative.directory/?search=react-native-tree-selection

JairajJangle: Hi @Simek,

Oh, you’re right! I think this library will cross the threshold in July 2025 (in about two months). I’ll revert this change for now and raise a PR later when the library is officially marked as unmaintained.

Thanks for pointing that out!

PR #1637, opened 2025-05-10
Add @rn-primitive components

dylmye:
# 📝 Why & how

Add all of the [rn-primitives](https://github.com/roninoss/rn-primitives) component packages. They are designed to be used with Expo. I also created an Expo Snack for each component.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -14790,7 +14790,8 @@
     "githubUrl": "https://github.com/roninoss/rn-primitives/tree/main/packages/accordion",
     "npmPkg": "@rn-primitives/accordion",
     "examples": [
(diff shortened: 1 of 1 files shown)


Simek (commented):
Hey @dylmye, thanks for adding new entries to the directory! 👍 

Before merging, let's remove `dev` flag from all packages besides `types`. According to the docs none of the packages should be installed as `devDependencies`, but looking at the code, it would be stripped, so we can leave the `dev` flag:
* https://rnprimitives.com/types/ (see "Manual installation")

dylmye:
Ah I wasn't aware it signified that, thank you!

PR #1657, opened 2025-05-21
Added my own fork of `react-native-copilot` so that Expo CLI does not keep warning me about it

Proman4713:
# 📝 Why & how
Add a new library

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -15425,6 +15425,7 @@
   },
   {
     "githubUrl": "https://github.com/Proman4713/react-native-copilot",
+    "npmPkg": "@proman4713/react-native-copilot",
     "ios": true,
     "android": true,
     "expoGo": true
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -15422,5 +15422,11 @@
     "ios": true,
     "android": true,
     "newArchitecture": true
-  }
+  },
+  {
+    "githubUrl": "https://github.com/Proman4713/react-native-copilot",
+    "ios": true,
+    "android": true,
+    "expoGo": true
+  },
 ]
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -15428,5 +15428,5 @@
     "ios": true,
     "android": true,
     "expoGo": true
-  },
+  }
 ]


Simek:
Hey @Proman4713, can you share the screenshot of warning you are wanting to resolve?

Simek:
Thanks for sharing! What's the exact command you are executing? Since we do the validation in CLI, in two places, and one more check is also made by `expo-doctor`.

Proman4713:
expo doctor

Proman4713:
i dont want to exclude these libraries from the directory check, why can't I just properly tell the directory whether or not they work for the new architecture?

Simek:
Usually it's easier to ignore private forks and packages than listing all of the forks of existing libraries. If we would do that, then the directory would be in tens of thousands entries, and the usability of it would be questionable due to shear amount of data, so we need to have some limiting factor.

You initial PR comment suggested directly that your package is not created to be used by others, it's just a fork for your own purpose, and repo stats reflect that. That's why I have suggested excluding the check. If your fork grow, and got some other users usage and traction, there is no reason why it cannot be added later in time.

Proman4713:
Alright, you have got a point.. I will manually exclude the fork on my end... But I'll still test the other packages and make PRs for them if they work

Simek:
Closing the PR according to the discussion above. 

Few notes on the remaining packages from the screenshot:
* `@fontawesome/fontawesome-svg-core` and `expo-checkbox` are currently listed in the directory
* `react-native-base64` package is no longer maintained, you should probably switch it to one of modern implementation, like https://github.com/craftzdog/react-native-quick-base64
* `react-native-vector-icons` is a legacy package, the new version split out into several per-icons-set packages, for the reference and migration steps see https://github.com/oblador/react-native-vector-icons/blob/master/MIGRATION.md
* `@react-navigation/material-bottom-tabs` has been moved in v6 to the `react-native-paper` package, see `react-navigation` changelog for the reference https://reactnavigation.org/docs/upgrading-from-6.x/#material-bottom-tab-navigator-now-lives-in-react-native-paper-package

Saying that, feel free to propose new additions, or missing packages in a new PRs, if you find any other warning for your codebase!

PR #1658, opened 2025-05-21
fix: change zod to monorepo

BrendanLouis1223:
# 📝 Why & how

[zod](https://github.com/colinhacks/zod) was being filtered as invalid. The package appears to be inside a monorepo so the githubUrl was updated to reflect that.

# ✅ Checklist


Simek (approved):
Hey @BrendanLouis1223, nice catch, thanks for the correction! 👌

PR #1659, opened 2025-05-22
AJ

ajaitech:


Simek:
Hey @ajaitech, thanks for attempting to add new entries, but this PR performs way more changes and additions than necessary, few highlights:
* we use scripts to format the entry list and applied formatting changes does not match the script output, so they will be reverted on the next run of it automatically
* `expo-facebook` is deprecated for few years in favor of `expo-auth-session`, and its source has been removed from expo/expo repository, hence it's not a valid entry anymore
* some of new entries you have propose are `devDependencies`, and if `expo` or `expo-doctor` warn you about them, it most likely means that they are listed in wrong dependency group in your project's `package.json`

Due to that, I'm going to close this PR for now, but feel free to follow up with set of new PR, one for each new entry, so it's easier to review them, and provide necessary feedback. 🙂

PR #1677, opened 2025-06-02
feat(directory): add hookform, cva and phosphor icons

mezotv:
# 📝 Why & how

Adds a few more packages im using in my app currently. Tested all of them on the new arch too!

- [x] Added libraries to **`react-native-libraries.json`**

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -15537,5 +15537,31 @@
     "android": true,
     "newArchitecture": true,
     "dev": true
+  },
+  {
+    "githubUrl": "https://github.com/react-hook-form/resolvers",
+    "npmPkg": "@hookform/resolvers",
+    "ios": true,
+    "android": true,
+    "expoGo": true,
+    "newArchitecture": true
+  },
+  {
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -15537,5 +15537,31 @@
...
+    "newArchitecture": true
+  },
+  {
+    "githubUrl": "https://github.com/joe-bell/cva/tree/main/packages/class-variance-authority",
+    "npmPkg": "class-variance-authority",
+    "ios": true,
+    "android": true,
+    "expoGo": true,
+    "newArchitecture": true,
+    "dev": true
```
Simek, line 15556:
```suggestion
    "newArchitecture": true
```

According to the docs this package should not be installed as development dependency:
* https://cva.style/docs/getting-started/installation

Simek (commented):
Hey @mezotv, thanks for adding new entries to the directory! 👍 

Just one thing to clear up before we can merge. 🙂

Simek (approved):
Thanks for the quick update! 👌

PR #1696, opened 2025-06-06
fix(better-auth): add missing npmPkg field

mezotv:
# 📝 Why & how
When I added the better-auth library I forgot to include the npmPkg field so this pr adds that

# ✅ Checklist

- [ ]  Added library to **`react-native-libraries.json`**

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -14555,6 +14555,7 @@
   },
   {
     "githubUrl": "https://github.com/better-auth/better-auth/tree/main/packages/better-auth",
+    "npmPkg": "better-auth",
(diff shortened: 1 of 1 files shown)


Simek:
Hey @mezotv, it should not be needed, since the package name matches repository name. This field is mostly for scoped packages, or ones in monorepo, see:
* https://github.com/react-native-community/directory

mezotv:
> Hey @mezotv, it should not be needed in this case, since the package name matches repository name. This field is mostly for scoped packages, or ones in monorepo, see:

Ahh, appreciate you clearing that up, thanks! I will close this then.

PR #1727, opened 2025-06-21
fix: preserve npm download data for scoped packages (fixes #1640)

JairajJangle:
# 📝 Why & how

Fixes #1640 - Download counts were not displaying for packages with `"@"` character (scoped packages) due to npm data being overwritten during bulk query merge. Scoped packages are processed individually but their npm data was being replaced with empty objects from the bulk results.
- Preserve existing npm data when merging bulk query results
- Ensures scoped packages retain their individually-fetched download stats
- Regular packages continue to work as before
---

<table>
  <tr>
    <td>before_fix</td>
    <td>after_fix</td>
  </tr>
  <tr>
    <td align="center">Before: download count is missing</td>
    <td align="center">After: download count is displayed</td>
  </tr>
</table>

---

# ✅ Checklist

- [x] Documented in this PR how to use the feature or replicate the bug.
- [x] Documented in this PR how you fixed or created the feature.

PR #1730, opened 2025-06-22
Updated assets/data.json to reflect #1727 fixes for development purposes

JairajJangle:
# 📝 Why & how

Updated `assets/data.json` file to reflect #1727 fixes for development purposes.

# ✅ Checklist

### DIFF
--- a/scripts/build-and-score-data.ts
+++ b/scripts/build-and-score-data.ts
@@ -130,6 +130,9 @@ async function buildAndScoreData() {
   data = data.map(project => ({
     ...project,
     npm: {
+      // Keep existing npm data (for scoped packages processed individually)
+      ...project.npm,
+      // Add bulk query results (for regular packages)
       ...(downloadsList.find(entry => entry.name === project.npmPkg)?.npm ?? {}),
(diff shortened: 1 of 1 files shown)


Simek:
Hey @JairajJangle, thanks for working on the Vercel Blob issue! 👍 

When it comes to the updating data JSON, usually this is done by maintainers, and only when we change the structure of entries, so the local dataset reflects the changes. What would be added by regeneration now is just missing download counts and several new entries, which should not hinder the development process.

The reason for this approach is that reviewing diff on that file is quite hard, and updating that file often clobbers the git history a bit. So let's revert the data changes for now, I will follow up that soon, with some scheduled changes, the rest looks good! 👌

JairajJangle:
Hi @Simek! Sure, no problem! If that’s the case, totally understandable. I’ll go ahead and revert the _assets/data.json_ changes and update the PR accordingly 👍

Thanks for the clarification! 🙌

Simek (approved):
Thanks for working on the issue, and all other updates in the process, great job! 💪

PR #1758, opened 2025-07-10
Add react-error-boundary to directory

Minori-ty:
# 📝 Why & how

add  [react-error-boundary](https://www.npmjs.com/package/react-error-boundary) to directory

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -16105,14 +16105,5 @@
     "web": true,
     "expoGo": true,
     "newArchitecture": true
-  },
-  {
-    "githubUrl": "https://github.com/expo/expo/tree/main/packages/eslint-config-expo",
-    "npmPkg": "eslint-config-expo",
-    "ios": true,
-    "android": true,
-    "web": true,
-    "expoGo": true,
-    "newArchitecture": true
   }
 ]
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -16096,5 +16096,23 @@
     "web": true,
     "expoGo": true,
     "newArchitecture": true
+  },
+  {
+    "githubUrl": "https://github.com/bvaughn/react-error-boundary",
+    "npmPkg": "react-error-boundary",
+    "ios": true,
+    "android": true,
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -16096,5 +16096,23 @@
...
+    "newArchitecture": true
+  },
+  {
+    "githubUrl": "https://github.com/expo/expo/tree/main/packages/eslint-config-expo",
+    "npmPkg": "eslint-config-expo",
+    "ios": true,
+    "android": true,
+    "web": true,
+    "expoGo": true,
+    "newArchitecture": true
```
Simek, line 16116: Hey @Minori-ty, in general the ESLint configs should be installed as developer dependency, so they should be omitted during different checks (made by Expo CLI or `expo-doctor`).

We also tend to not accept ESLint related packages, since there is so many of configs and plugins, and they will clutter the directory.

Minori-ty: Thank you for your reply. I will try to ignore eslint-config-expo in expo-doctor.

> Hey @Minori-ty, in general the ESLint configs should be installed as developer dependency, so they should be omitted during different checks (made by Expo CLI or `expo-doctor`).
> 
> We also tend to not accept ESLint related packages, since there is so many of configs and plugins, and they will clutter the directory.

Thank you for your reply. I will try to ignore eslint-config-expo in expo-doctor.

Simek: We should not be checking for `devDependencies` in doctor, if we do, that's a bug.

You can also ignore the given packages via `doctor` config, see:
* https://docs.expo.dev/versions/latest/config/package-json/#doctor

Minori-ty: > We should not be checking for `devDependencies` in doctor, if we do, that's a bug, and please let us know!
> 
> You can also ignore the given packages via `doctor` config, see:
> 
> * https://docs.expo.dev/versions/latest/config/package-json/#doctor

Thank you so much for telling me about this config. I'll go try it right away.

PR #1797, opened 2025-08-08
added expo-stack-bottom-sheet

efstathiosntonas:
@Simek no npm package, just a "demonstration", shall we keep it or ditch it? It looks nice 😄 

# 📝 Why & how

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -16453,5 +16453,11 @@
     "ios": true,
     "android": true,
     "newArchitecture": true
+  },
+    {
+    "githubUrl": "https://github.com/rit3zh/expo-stack-bottom-sheet",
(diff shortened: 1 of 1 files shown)


Simek:
Hey @efstathiosntonas, we do not support stand-alone demos at this moment, since we don't want to mix them on the list with installable packages. One thing which we can do now i to list that repository as an example in one of the packages, i.e. Reanimated.

However, in a near future I will think about this, and discuss the idea of creating separate list of demonstration apps with other stakeholders, maybe this would be a nice addition.

efstathiosntonas:
I ll add it under reanimated examples later today. 

I believe demonstration apps should only be targeted demos like this one, not full blown apps, just saying.

PR #1815, opened 2025-08-12
Add `react-native-bundle-discovery`

retyui:
# 📝 Why & how

Add `react-native-bundle-discovery` package

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -16590,5 +16590,13 @@
     "ios": true,
     "android": true,
     "newArchitecture": true
+  },
+  {
+    "githubUrl": "https://github.com/retyui/react-native-bundle-discovery",
+    "images": [
+      "https://raw.githubusercontent.com/retyui/react-native-bundle-discovery/main/assets/img.png"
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -16590,5 +16590,13 @@
     "ios": true,
     "android": true,
     "newArchitecture": true
+  },
+  {
+    "githubUrl": "https://github.com/retyui/react-native-bundle-discovery",
+    "images": [
+      "https://raw.githubusercontent.com/retyui/react-native-bundle-discovery/main/assets/img.png"
+    ],
+    "alternatives": ["react-native-bundle-visualizer", "expo-atlas-without-expo", "expo-atlas"],
```
Simek, line 16599: Hey @retyui, thanks for the contribution! 👍 

Just one thing, we use the `alternatives` field only to recommend library over others if ones on the list are no longer maintained, or does not support New Architecture, see:

I don't think it's true for any of the libraries in the current list.


```suggestion
```

PR #1829, opened 2025-08-14
Update data for `react-native-google-cast` and `jwt-decode`

JustJoostNL:
# 📝 Why & how
`react-native-google-cast` & `jwt-decode` both support the new architecture. `jwt-decode` always did and `react-native-google-cast` has been updated recently: https://github.com/react-native-google-cast/react-native-google-cast/releases/tag/4.9.0. I also tested both libraries succesfully in my app.

# ✅ Checklist


Simek (approved):
Thanks for the context and updates! 🚀

PR #1849, opened 2025-08-24
feat: add react-native-blossom-ui libraries

deepakkumardk:
# 📝 Why & how

Added a new react-native-blossom-ui monorepo libraries

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -16749,5 +16749,45 @@
     "ios": true,
     "android": true,
     "newArchitecture": true
+  },
+  {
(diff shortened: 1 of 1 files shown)


Simek (commented):
Hello @deepakkumardk, thanks for adding new entries to the directory! 👍 

I have a question about `@react-native-blossom-ui/showcase` - this package looks like demo app to showcase the components, not the dependency which should be installed in the users projects. Can you provide a bit more context on that package? If my assessment is correct, let's skip that entry.

deepakkumardk:
Yes, I can remove the showcase package that's just for internal use in the demo project.

Simek (approved):
Thanks for more context on that, and the quick update! 🚀

PR #1861, opened 2025-08-31
update Legend List and State with full support

jmeistrich:
# 📝 Why & how

Updates the support of Legend List and Legend State

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -14119,8 +14119,13 @@
     "android": true,
     "ios": true,
     "macos": true,
+    "windows": true,
     "expoGo": true,
-    "newArchitecture": true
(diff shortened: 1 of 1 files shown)


jmeistrich:
Fixed the conflicts, not sure why it had started the PR from such an old base...

I am making new docs which lists more supported platforms but I haven't published it yet. I have verified that it works on FireTV with a company's internal product, but I don't have a public repo with it. I've been told that it works well in the other platforms but have not yet tested myself. It's all javascript just using ScrollView, so there's no reason why it wouldn't work.

Should I make public repos for each of the platforms before it can be listed as supported here? I don't have the devices so I can't really fully test it myself 😅

PR #1900, opened 2025-09-24
docs(library): adding @flixsrota/player to the directory

NIKHIL0VERMA:
# 📝 Why & how
This PR include a new library called @flixsrota/player an alternative player to show youtube videos without bloats, share buttons and recommendation junk.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -17097,13 +17097,6 @@
     ],
     "ios": true,
     "android": true,
-    "web": false,
-    "expoGo": true,
-    "unmaintained": false,
-    "dev": false,
-    "template": false,
-    "alternatives":[
-      "react-native-youtube-iframe"
-    ]
+    "expoGo": true
   }
 ]


react-native-libraries.json
```diff
@@ -17082,5 +17082,28 @@
...
+    "ios": true,
+    "android": true,
+    "web": false,
+    "expoGo": true,
+    "unmaintained": false,
+    "dev": false,
+    "template": false,
+    "alternatives":[
+      "react-native-youtube-iframe"
+    ]
```
Simek, line 17107: Hey @NIKHIL0VERMA, thanks for adding a new entry to the directory! 👍 

Just two small suggestion before we can ship it:
* `alternatives` list only applies to libraries which are not maintained anymore, or do not support New Architecture, `react-native-youtube-iframe` seems to do not qualify for those criterions, more details:
* flags with `false` values can be safely skipped

```suggestion
    "expoGo": true
```

PR #1906, opened 2025-09-29
feat: Add react-native-unarchive

pushpender-singh-ap:
# 📝 Why & how

This PR adds react-native-unarchive library for React Native developers that need to extract RAR, ZIP, CBR and CBZ files in their apps, this package is a useful resource. It supports React Native's new Turbo Module architecture and was created in Kotlin and Objective-C++.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -9694,13 +9694,6 @@
     "android": true
   },
   {
-    "githubUrl": "https://github.com/pushpender-singh-ap/react-native-unarchive",
-    "npmPkg": "react-native-unarchive",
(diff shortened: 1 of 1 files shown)


pushpender-singh-ap:
I don't know why action is fail here is the npm link for more details: https://www.npmjs.com/package/react-native-unarchive

pushpender-singh-ap:
@Simek , Sure, I’ll move this entry to the end of the file. I initially thought it would be better to keep all libraries from the same author grouped together, but no problem—I’ll make the changes.

PR #1933, opened 2025-10-14
Updated react-native-play-age-range-declaration package name

Gautham495:
Updated library name from react-native-play-age-signals to react-native-play-age-range-declaration manage both Android & iOS Play Age Signal and Declared Age APIs.

Before it was only for Android and now it is for both.

It now uses Nitro Modules from Turbo Modules and I removed all of the Obj C code as I was having build issues.

- [ ] Documented in this PR how you fixed an issue or created the feature.

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -17286,9 +17286,10 @@
     "ios": true
   },
   {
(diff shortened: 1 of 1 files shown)


Simek (approved):
Hey @Gautham495, thanks for the update! 🚀

PR #1949, opened 2025-10-20
Add Expo config plugins support

WookieFPV:
# 📝 Why & how
Currently its not easy to know which packages include an expo config plugin.
Some packages even don't have config plugins, some have a build in config plugin and some need 3rd Party config plugins.

This PR adds automatic detection if the npmPkg has a config plugin. (Checks if file `app.plugin.js` is in the repo).
Third Party config plugins are supported by using a url to the 3rd Party package.
I added the 3rd Party Plugins from using the [Expo 3rd Party List](https://github.com/expo/config-plugins/tree/main/packages).

# ✅ Checklist

- [X] Documented in this PR how to use the feature.
- [X] Documented in this PR how you created the feature.

### DIFF
--- a/types/index.ts
+++ b/types/index.ts
@@ -33,7 +33,7 @@ export type Query = {
   hasImage?: string;
   hasTypes?: string;
   hasNativeCode?: string;
-  configPlugin?: boolean | string;
+  configPlugin?: string;
   isMaintained?: string;
   isPopular?: string;
   isRecommended?: string;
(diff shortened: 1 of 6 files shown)


types/index.ts
```diff
@@ -33,6 +33,7 @@ export type Query = {
   hasImage?: string;
   hasTypes?: string;
   hasNativeCode?: string;
+  configPlugin?: boolean | string;
```
Simek:
```suggestion
  configPlugin?: string;
```

Query can only be string.

WookieFPV:
> Hey @WookieFPV, thanks so much for working on that feature! 🔥
> 
> I have left few small suggestions and nits to address before we will merge, but in general it looks and functions great, good job! 👍

Hey @Simek 
Thanks for the fast review and great improvements 🚀 
I added those Improvements are some minor wording improvements

PR #1980, opened 2025-11-05
chore: added react-native-speech-recognition-kit

Gautham495:
Added a library called react-native-speech-recognition-kit as there is no available speech to text (STT) libraries available in react native which runs on Bare React Native.

This library works on both Expo and Bare as it does not use Expo Modules and is built with Turbo Modules.

- [ ] Documented in this PR how you fixed an issue or created the feature.

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -18078,5 +18078,14 @@
     "ios": true,
     "android": true,
     "newArchitecture": true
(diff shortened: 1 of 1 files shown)


Simek (approved):
Hello @Gautham495, thanks for adding a new entry to the directory! 👍

PR #2004, opened 2025-11-19
Add a script to generate `llms.txt`

amandeepmittal:
# 📝 Why & how

@kadikraman requested that a script be added to reactnative.directory to generate an `llms.txt` file. Once hosted, this file can be reused in AI integration tools and AI docs integrations such as [Expo documentation](https://docs.expo.dev/).

- This PR adds a new script (`bun run `llms:generate`) to generate the `llms.txt` file.
- It uses the curated dataset from `react-native-libraries.json` plus the cached GitHub metadata from `assets/data.json`.
- The output lists each library with name, repo link, description, supported platforms, and New Architecture status.
- Updates README.md with a " How do I generate `llms.txt`?" that provides instructions on how to generate this file.

Example:

CleanShot 2025-11-19 at 19 38 40@2x

## Test plan

- Run `bun run generate:llms`
- Run `bun run start`
- Visit http://localhost:3000/llms.txt in your browser.


kadikraman:
Hi @Simek! The purpose of this PR is to make React Native libraries easier to discover via LLMs, including the docs AI search. The RN Directory is not easily scannable with LLMs so we were finding that our docs search was giving incorrect information regarding RN libraries, or not even knowing they exist. With this, we can have more accurate and up to date information as well as be able to link back to RN Directory as the source of the info.

What do you think?

scripts/generate-llms.ts
```diff
@@ -0,0 +1,145 @@
...
+  fireos: 'Fire OS',
+  horizon: 'Meta Horizon',
+  ios: 'iOS',
+  macos: 'macOS',
+  tvos: 'tvOS',
+  visionos: 'visionOS',
+  vegaos: 'Vega OS',
+  web: 'Web',
+  windows: 'Windows',
+} as const;
```
Simek: Let's retain the spit between platforms and compatibility terms, and reuse existing helpers from:
* https://github.com/react-native-community/directory/blob/main/components/Filters/helpers.ts

We can also move them to global `/utils` if needed.

PR #2097, opened 2026-01-07
fix search input desync with debounced query

### LINKED ISSUE #2096: 'Search libraris' input field clears text while still typing
In the main page in this [reactnative.directory](reactnative.directory) website, the text input field used to search and filter libraries is clearing typed text while I am still typing.

I typed something like 'whis' and suddenly it deletes all of the text inside the input and updates the url query params `?search=whis`. 

I noticed that this issue happens frequently when the list refreshes.

https://github.com/user-attachments/assets/d84cfeaa-a429-49c6-8c2a-dc71dcaa0540

Simek:
# 📝 Why & how

Hopefully fixes #2096

Looks like when clearing a search input and retyping new value quickly within the debounce window (200ms) it might sometimes lead to query param and input value desync, due to `search` induced `useEffect` triggering before debounce ended. 

The `useEffect` block has been added few months ago due to desync caused by pressing homepage logo link, but after changes to  link element within Tailwind refactor this should no longer be a case, which made this update effect redundant.

# ✅ Checklist

# Preview

https://react-native-directory-q94s4znsj-rndir.vercel.app/

### DIFF
--- a/components/Search.tsx
+++ b/components/Search.tsx
@@ -43,12 +43,6 @@ export default function Search({ query, total }: Props) {
     }
   }, [isApple]);
 
-  useEffect(() => {
-    if (inputRef?.current && !search) {
-      inputRef.current.clear();
-    }
-  }, [search]);
-
   const typingCallback = useDebouncedCallback((text: string) => {
     void replace(urlWithQuery('/', { ...query, search: text, offset: null }));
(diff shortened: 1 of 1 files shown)


ahmedshayea (commented):
That resolved the issue entirely.

PR #2332, opened 2026-03-30
Fix packages search input on mobile

cornejobarraza:
# 📝 Why & how

The search input in the packages page has some overlapping issues with the magnifying glass icon.

To fix this I just matched the `<Search />` component style to that of `<QuickSearch />`:

| Before                                                                                                                                     | After                                                                                                                                     |
| ------------------------------------------------------------------------------------------------------------------------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------- |
| input_before | input_after |

I also updated the `id` for the quick search input since it was using the same value in both places.

# ✅ Checklist

### DIFF
--- a/components/QuickSearch.tsx
+++ b/components/QuickSearch.tsx
@@ -144,7 +144,7 @@ export default function QuickSearch({ style }: Props) {
             </View>
             <TextInput
               ref={inputRef}
-              id="search"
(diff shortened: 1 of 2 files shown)


Simek (approved):
Nice catch, thanks for the fix! 🚀

PR #2385, opened 2026-04-19
update @maplibre/maplibre-react-native github url

KiwiKilian:
# 📝 Why & how

It looks like the check, if a README.md is present is done from the GitHub URL. We have placed the README at the root and copy it into the package only during packaging. So we should probably stick with this root URL or the check should see if there is a README at the given GitHub URL **or** within the published package? Not sure. But for now I don't wont to give up score for this 😄.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -10178,7 +10178,7 @@
     "android": true
   },
   {
-    "githubUrl": "https://github.com/maplibre/maplibre-react-native/tree/main/package",
+    "githubUrl": "https://github.com/maplibre/maplibre-react-native",
     "npmPkg": "@maplibre/maplibre-react-native",
     "examples": [
       "https://github.com/maplibre/maplibre-react-native/tree/main/examples/expo-app",


KiwiKilian:
@Simek hmmm – your validation wants us to point to the package directory... But we want the main README at the root, because that's were most traffic enters... Do you have good examples of other repositories handling this?

Simek:
Hey @KiwiKilian,

The README is visible on the package page since we fetch it from the package bundle:
* https://reactnative.directory/package/@maplibre/maplibre-react-native

The scoring logic checks GitHub repository and npm registry for REDME file existence, but there might be an issue with the logic in there, will look at this soon.

> Do you have good examples of other repositories handling this?

As a workaround, you can address this on your side by creating symlink in package directory to the README file at the repository root (via regular symlink or custom GitHub one). Example:
* https://github.com/rrousselGit/riverpod/tree/master/packages/riverpod

Simek:
Checked the responses, and it looks like a bug in the npm registry API, which besides the fact that bundle includes README, and it's displayed on the registry package page returns error (no README data found) in the response:
* https://registry.npmjs.org/@maplibre/maplibre-react-native

Screenshot 2026-04-19 104009

On the Directory side we also fetch GitHub root file list, so we can detect README existence in there, but there is no guarantee that it would be copied over, without checking the publish scripts which is out of our scope, so for now I would recommend using a symlink. 

You can try to report this API issue to npm (via on-site contact, since Issues are disable on the companion GitHub repo), but chances that they will address that are unfortunately quite slim.

KiwiKilian:
Thanks for looking into this! I will check, if the symlink works. How fast is RN directory updated?

But as the README content is displayed on RN directory, wouldn't it be possibel to add this as an alternative check for the existence of the README? Or the existence of the README in the file tree in the "code" section?

Simek:
> But as the README content is displayed on RN directory, wouldn't it be possibel to add this as an alternative check for the existence of the README? Or the existence of the README in the file tree in the "code" section?

Package READMEs are fetch client side, while libraries data exposed by API is fetch server side and refreshed on each deployment.

I have checked the data for similar cases and there are 87 package with both README files set to `false`, and for majority of them assessment is true. There are less than 20 packages which shares the similar issue like `maplibre`, we can probably add an additional bundle fetch to confirm the correctness of the data during the deployment, but will nee to measure first how it impacts the deployment time, and if there are any rate limits on unpkg API which could cause some problems.

Can you create an issue for that case in the repo, so we can track the progress in there, instead of this PR?

KiwiKilian:
Done, thanks for your investigations!

#2389

PR #2400, opened 2026-04-22
Fix bookmark button on mobile

cornejobarraza:
# 📝 Why & how

There is an overlap between the GitHub and Bookmark buttons on mobile.

I fixed this by rendering separate `<BookmarkButton />` components depending on screen size:

| Before                                                                                                                                               | After                                                                                                                                               |
| ---------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
| mobile_bookmark_before | mobile_bookmark_after |

# ✅ Checklist

- [x] Documented how you found or replicated the issue.
- [x] Explained how you fixed the issue or built the feature.
- [x] Described how to use or verify the change.

PR #2406, opened 2026-04-23
Fix updated at on mobile

cornejobarraza:
# 📝 Why & how

Missed some spacing issues with `<UpdatedAtView />`  when fixing the bookmark button on PR https://github.com/react-native-community/directory/pull/2400:

| Before                                                                                                                                       | After                                                                                                                                       |
| -------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------- |
| popular_before | popular_after |

I also fixed a typo in the component file itself, it was named `UpdateAtView.tsx` instead of `UpdatedAtView.tsx`

# ✅ Checklist


Simek:
Hey thanks for the follow up! 

Just tested the change locally, and it looks like the change did not fix the issue on the trending page (in desktop view):

Screenshot 2026-04-23 231940

components/Library/index.tsx
```diff
@@ -77,7 +77,11 @@ export default function Library({
           </View>
         )}
         {showTrendingMark && library.popularity && (
-          <View style={tw`mb-1 flex-row items-center justify-between gap-6 pr-8`}>
+          <View
+            style={[
+              tw`mb-3 flex-row items-center justify-between gap-6`,
+              !isSmallScreen && tw`pr-10`,
```
Simek: Also, let's retain the original spacing - `8` instead of `10`.

cornejobarraza:
Just pushed a change that should take care of that and restore the original padding value

Simek (approved):
Thanks for the follow-up update, and for the name correction! 👌

PR #2416, opened 2026-04-29
Parallelize SSR fetches on the home page

azizbecha:
## Summary
- `pages/index.tsx` was awaiting five `ssrFetch` calls one after another inside `getServerSideProps`, then awaiting each `.json()` sequentially. The fetches don't depend on each other.
- Fan them out with `Promise.all` for both the requests and the JSON parses, so the waterfall collapses into a single roundtrip's worth of latency on the home page.

## Test plan
- [x] `bunx oxlint pages/index.tsx` — clean
- [x] `bunx oxfmt --check pages/index.tsx` — clean
- [x] `bun start` and load `/` locally; dev-server log shows the five `/api/libraries...` requests issued in the same window (~470ms each, overlapping) instead of stacked sequentially. Warm renders return in ~24-185ms with all five API calls completing in ~5-15ms each, in parallel. No runtime errors.

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -22,47 +22,58 @@ export async function getServerSideProps(ctx: GetServerSidePropsContext) {
     };
   }
 
-  const mostDownloadedResponse = await ssrFetch(
-    '/libraries',
-    {
-      order: 'downloads',
-      limit: LIMIT.toString(),
-      isMaintained: 'true',
-      hasNativeCode: 'true',
-    },
-    ctx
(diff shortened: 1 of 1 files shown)


Simek (approved):
Hey @azizbecha, thanks nice catch and improvement! 👌 

We already use parallel promises for package pages fetches, but I have missed this on the home page. 

Would you be interested in optimizing also the Popular page fetches? Can be done in a separate PR.

PR #2418, opened 2026-04-29
Parallelize SSR fetches on the Popular page

azizbecha:
## Summary
- Follow-up to #2416 as suggested in review.
- `pages/popular.tsx` was awaiting twelve `ssrFetch` calls sequentially in `getInitialProps` (one per platform), then awaiting each `.json()` one at a time. The fetches don't depend on each other.
- Fan them out with `Promise.all` for both the requests and the JSON parses.

## Test plan
- [x] `bunx oxlint pages/popular.tsx` — clean
- [x] `bunx oxfmt --check pages/popular.tsx` — clean
- [x] `bun start` and load `/popular` locally; dev-server log shows all twelve `/api/libraries?...` requests issued in the same window instead of stacked sequentially. Warm renders return in ~26ms with all twelve API calls completing in ~15ms each, in parallel. No runtime errors.

### DIFF
--- a/pages/popular.tsx
+++ b/pages/popular.tsx
@@ -15,48 +15,63 @@ const DEFAULT_QUERY = {
 };
 
 PopularPage.getInitialProps = async (ctx: NextPageContext) => {
-  const coreResponse = await ssrFetch(
-    '/libraries',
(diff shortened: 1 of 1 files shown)


Simek (approved):
Thanks for the follow up, LGTM! 🚀

PR #2430, opened 2026-05-02
fix contradictory ARIA on Button

azizbecha:
## Summary

The non-link `Button` variant in `components/Button.tsx` set `role=\"button\"` together with `accessible={false}` on the same `View`. These cancel out: `accessible={false}` hides the element from assistive technology, so the explicit `role=\"button\"` never reaches screen readers.

This drops `accessible={false}` from that `View` so the role is actually announced. The link variant is unchanged — there the inner `View` is correctly hidden because the wrapping `<A>` is the accessible element.

### DIFF
--- a/components/Button.tsx
+++ b/components/Button.tsx
@@ -48,7 +48,7 @@ export function Button({
           </View>
         </A>
       ) : (
-        <View role="button" focusable={false} style={buttonStyle} accessible={false} {...rest}>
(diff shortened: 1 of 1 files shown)


Simek (approved):
Nice catch @azizbecha! 👍 

This prop was introduced to prevent double element focus in certain cases, but looks like `focusable` prop is enough to avoid that.

PR #2431, opened 2026-05-02
add rel="noopener noreferrer" to external links

azizbecha:
## Summary

The shared `<A>` component in `common/styleguide.tsx` defaults `target` to `_blank` but does not emit a matching `rel` attribute. Links opened in a new tab without `rel="noopener"` allow the new page to access `window.opener`, which is a known tabnabbing vector, and without `noreferrer` the referrer is leaked to the destination.

This sets `rel="noopener noreferrer"` (on both the web `rel` prop and the React Native `hrefAttrs`) whenever the resolved target is `_blank`. Internal `target="_self"` links are unaffected.

### DIFF
--- a/common/styleguide.tsx
+++ b/common/styleguide.tsx
@@ -140,7 +140,11 @@ export function A({
         href={href}
         numberOfLines={containerStyle ? 1 : undefined}
         target={target ?? '_blank'}
-        hrefAttrs={{ target: target ?? '_blank' }}
+        rel={(target ?? '_blank') === '_blank' ? 'noopener noreferrer' : undefined}
+        hrefAttrs={{
+          target: target ?? '_blank',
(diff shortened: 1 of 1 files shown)


common/styleguide.tsx
```diff
@@ -140,7 +140,11 @@ export function A({
         href={href}
         numberOfLines={containerStyle ? 1 : undefined}
         target={target ?? '_blank'}
-        hrefAttrs={{ target: target ?? '_blank' }}
+        rel={(target ?? '_blank') === '_blank' ? 'noopener noreferrer' : undefined}
```
Simek, line 143: Can we extract `target ?? '_blank'` to a variable to avoid checks repetition, and for a bit more code clarity?

Simek:
Due to lack of activity in last month I have made a separate PR addressing that (which is linked above).

Thanks for bringing this up tho! :)

PR #2432, opened 2026-05-02
set descriptive alt text on library screenshot thumbnails

azizbecha:
## Summary

`components/Library/Thumbnail.tsx` rendered the screenshot `` with `alt=""`, which signals to assistive technology that the image is purely decorative. These are actual package screenshots, so the empty alt hides meaningful content from screen readers.

This threads the library name (`npmPkg ?? github.fullName`) from `Library/index.tsx` into `Thumbnail` and uses it as the alt text (e.g. `"react-native-foo screenshot"`).

### DIFF
--- a/components/Library/Thumbnail.tsx
+++ b/components/Library/Thumbnail.tsx
@@ -7,11 +7,12 @@ import tw from '~/util/tailwind';
 
 type Props = {
   url: string;
+  name: string;
 };
 
 const GITHUB_PREVIEW_MIN_WIDTH = 640;
 
-function Thumbnail({ url }: Props) {
+function Thumbnail({ url, name }: Props) {
   const { width, height } = useWindowDimensions();
 
   const [isLoaded, setLoaded] = useState(false);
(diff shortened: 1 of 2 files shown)


components/Library/Thumbnail.tsx
```diff
@@ -56,7 +57,7 @@ function Thumbnail({ url }: Props) {
             <img
               src={url}
               onLoad={() => setLoaded(true)}
-              alt=""
+              alt={`${name} screenshot`}
```
Simek, line 60: In general from a11y/SEO perspective image descriptions should be unique/distinctive. We cannot set proper descriptions since we does not have clue what the screenshot includes, but let's at least use the index in the sentence to number the images.

Also, let's pass `alt` as a prop, and avoid using package specifc `name` prop, so component would be a bit more generalized.

Simek:
Thanks for the fix proposition! 👍 Closing in favour of mentioned above PR.

PR #2465, opened 2026-05-16
Add munim-bluetooth to the directory

sheehanmunim:
# Why & how

This PR adds `munim-bluetooth` (https://github.com/munimtechnologies/munim-bluetooth) to the directory.

The package supports iOS and Android, includes an example app, ships an Expo config plugin, and is built on Nitro Modules, so it is marked as New Architecture only.

# Checklist

- [x] Added library to `react-native-libraries.json`


Simek (approved):
Hey @sheehanmunim, thanks for adding new entry to the directory! 👍

PR #2506, opened 2026-06-10
Update previous and Add new nitro-speech library

NotGeorgeMessier:
# 📝 Why & how

I'm renaming `@gmessier/nitro-speech` into `react-native-nitro-speech` as it has grown sufficiently and shown stability

I want to add `react-native-nitro-speech` to the list and mark previous version as "not actively maintained"

I will maintain both versions for the next 1-2 releases and then focus only on the new label

The API is identical, only the package name is new

Also: examples were added

# ✅ Checklist


Simek (approved):
Hey @NotGeorgeMessier, thanks for the update and adding new entry! 👍 

I have removed the MD examples since we only allow full projects/Snacks in there, see:
* https://github.com/react-native-community/directory#examples

PR #2511, opened 2026-06-10
Set "newArchitecture": true for stripe-terminal-react-native

timlin-stripe:
Set "newArchitecture": true for stripe-terminal-react-native

# 📝 Why & how

The stripe-terminal-react-native currently is rely on the interop layer but we've test on every release with test-app which is newArchitecture enabled. So we'd like to set the newArchitecture flag to true if possible.  

# ✅ Checklist


Simek (approved):
Hey @timlin-stripe, thanks for the context and update! 👍

PR #2546, opened 2026-06-20
add Fire OS + Vega OS compatibility for 27 verified libraries

mosesroth:
# 📝 Why & how

Add Fire OS (`fireos: true`) and/or Vega OS (`vegaos: true`) for 27 libraries.

23 UI libraries were built and run on the target platform and confirmed to render the library's actual UI (not a blank/error screen). On Vega these render through the already-ported native modules via the Module Resolver Preset, so `vegaos: true` (no dedicated `@amazon-devices/*` package).

4 libraries are pure-JS / behavioral (color normalization, clipboard, query persistence, PDF generation) verified by exercising the real API and asserting its returned value; these are engine-identical across platforms.

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -1285,7 +1285,9 @@
     "ios": true,
     "android": true,
     "expoGo": true,
-    "npmPkg": "react-native-calendar-picker"
+    "npmPkg": "react-native-calendar-picker",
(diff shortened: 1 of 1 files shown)


Simek (approved):
Thanks for the updates and so many contributions lately! 🙂

PR #2547, opened 2026-06-20
add Fire OS + Vega OS compatibility for 30 verified libraries

mosesroth:
# 📝 Why & how

Add Fire OS (`fireos: true`) and/or Vega OS (`vegaos: true`) for 30 libraries.

8 are pure-JS / behavioral (semver, date-fns, @tanstack/react-query, @reduxjs/toolkit, xstate, @google/generative-ai, expo, react-compiler-runtime) verified by exercising the real API and asserting its returned value; these are engine-identical across platforms.

22 are UI libraries built and run on the target platform and confirmed to render the library's actual UI (not a blank/error screen).

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -648,7 +648,9 @@
     "githubUrl": "https://github.com/24ark/react-native-step-indicator",
     "ios": true,
     "android": true,
-    "expoGo": true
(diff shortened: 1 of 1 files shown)


Simek (approved):
Thanks for the updates! 🚀

PR #2566, opened 2026-06-24
Add token-based matching to improve search recall for non-contiguous package names

hryhoriiK97:
# 📝 Why & how

Improves search recall for libraries whose names contain all the query words but not as a contiguous substring.

**Problem:** Searching for `react-native-markdown` would not surface libraries like `react-native-enriched-markdown` or `react-native-live-markdown` in the `cleanedUpNameMatchPoints`, `npmPkgNameMatchPoints`, and `repoNameMatchPoints` scoring tiers, because those tiers only checked for contiguous substring matches. The word `enriched`/`live` in the middle broke the match entirely, scoring 0.

**Solution:** After stripping the `react-native`/`react` prefix from both the query and package name, split into tokens and check if every token appears in the cleaned-up name. This adds a fallback scoring tier to three match checks:

- `npmPkgNameMatchPoints`: 125 (contiguous) → **100 (token match)** → 0
- `cleanedUpNameMatchPoints`: 100 (contiguous) → **75 (token match)** → 0
- `repoNameMatchPoints`: 50 (contiguous) → **35 (token match)** → 0

Contiguous substring matches still score higher, so a library like `react-native-markdown-renderer` correctly ranks above `react-native-enriched-markdown` for the query `react-native-markdown` — but both now appear in results instead of only the former.

https://github.com/user-attachments/assets/0d3d8cdb-b406-4cfa-af08-2acbe78a7db2

# ✅ Checklist

### DIFF
--- a/util/search.ts
+++ b/util/search.ts
@@ -25,13 +25,6 @@ function calculateMatchScore(
       ? 150
       : 0;
 
-  const npmPkgNameMatchPoints =
-    !isEmptyOrNull(npmPkg) &&
-    (npmPkg.includes(querySearch) ||
-      npmPkg.replaceAll(NPM_NAME_CLEANUP_REGEX, ' ').toLowerCase().includes(querySearch))
-      ? 125
-      : 0;
-
   const cleanedUpName = npmPkg
     .replace('react-native', '')
     .replace('react', '')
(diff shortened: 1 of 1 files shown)


Simek (approved):
Tested the changes locally and it looks like nice improvement, thanks! 🙏 

This approach also makes single word queries more accurate which is a nice side-effect. 🙂

PR #2582, opened 2026-06-27
fix: retry npm downloads API on 429 status (rate-limit retry never fired)

mosesroth:
## Problem

`scripts/fetch-npm-download-data.ts` has a retry path whose own log messages describe it as handling the npm rate limit ("Looks like we have reached the npm API rate limit!", "Retrying fetch for ..."). That retry lives in the `catch` block, which only runs when `fetch` throws. The npm downloads API signals rate limiting with an HTTP 429 status, and `fetch` does not throw on a 429. So a 429 hits the `if (response.status !== 200)` branch first and returns `{ npm: null }`, and the retry never runs.

The effect shows up in `scripts/validate-new-entries.ts` (`bun ci:validate`), which makes one downloads call per changed entry. A `npm: null` result fails that entry's check, so a PR touching enough entries to brush the rate limit fails CI on a transient 429 with no recovery. The limit is a rolling window that hits a different subset each time, so re-running fails on a different entry instead of passing.

## Evidence

I replayed `validate-new-entries`'s request pacing (batches of 5, 1s stagger, 3s between batches) against the real `api.npmjs.org/downloads` endpoint. The rate limit comes back as a 429 status every time: a 60-request burst returned 60 responses with status 429 and 0 thrown errors. So these responses currently take the no-retry status branch rather than the `catch`.

## Fix

Handle 429 (and transient 5xx) with the same `REQUEST_SLEEP` backoff and `ATTEMPTS_LIMIT` retry the `catch` already uses, before falling through to the error return. Permanent failures (404, malformed body, retries exhausted) still return `{ npm: null }` as before.

The change is about 16 lines, reuses the existing retry code, adds no dependencies, and passes `oxlint --type-aware && oxfmt --check`.

### DIFF
--- a/scripts/fetch-npm-download-data.ts
+++ b/scripts/fetch-npm-download-data.ts
@@ -18,6 +18,18 @@ export async function fetchNpmDownloadData(pkgData: LibraryType, attemptsCount =
     const response = await fetch(url);
 
     if (response.status !== 200) {
+      // Rate limiting comes back as a 429 status (and transient outages as 5xx).
+      // fetch does not throw on these, so without this they skip the catch block
+      // below where the retry lives, and a single 429 fails the entry. Retry them
+      // with the same backoff the catch uses.
+      if ((response.status === 429 || response.status >= 500) && attemptsCount < ATTEMPTS_LIMIT) {
+        await sleep(REQUEST_SLEEP, REQUEST_SLEEP * 2);
+        console.log(
+          `⬇️ [NPM DOWNLOADS API] status ${response.status} for ${npmPkg}, retrying (${attemptsCount + 1})`
+        );
+        return await fetchNpmDownloadData(pkgData, attemptsCount + 1);
+      }
+
       console.error(
         `⬇️ [NPM DOWNLOADS API] npm API has returned invalid response - status ${response.status}!`
       );


Simek:
Hey, 

yeah, when writing the `validate-new-entries` script I did not anticipate it being used with more than like 10 entries, so I have used the npm fetcher directly. The best approach would be to extract the batching and sequantial fetch helpers from `build-and-score-data` script and fetch npm download in batches also in `validate-new-entries` script:
* https://github.com/react-native-community/directory/blob/main/scripts/build-and-score-data.ts#L463

The request batching is configured in main data fetch script in a way which prevents rate limiting to be engaged, and stats for all 2,5k+ entries can be fetched with no issue. I think it's better to avoid 429 responses at all instead of handling them directly, since after the first 429 response the `REQUEST_SLEEP` time might not be enough, since in so past test I have made the IP can be blacklisted for up to a minute, and retrying fetches only increase that time.

Another thing is that main script uses https://npm-stat.com/ as a download counts source, and npm dowload API is just used as fallback (in the past it was main source). The idea for npm check in `validate-new-entries` is mostly to verify if package is published and publicly avaible, so we also might switch the endpoint to npm registry API instead.

Simek:
Just updated the validate script to check packages in sequence rather than in parallel batches. Modified 30+ entries locally and made sure that rate limit does not trigger. 

Please let me know if this works for you case! Happy to iterate on this further, if it's not.

mosesroth:
Hi Bartosz,

That's actually a better fix, thanks! I'll close this PR.

Moses

PR #2586, opened 2026-06-27
fix: allow legitimately-private single-package repos past the isPrivate check (#2584)

### LINKED ISSUE #2584: Validation rejects single-package repos that set `private: true` at the package root
## What happens

`validate-new-entries.ts` rejects any added or updated entry whose GitHub `package.json` has `private: true`:

```
Extracted 'package.json' from <url> is marked as private! You might be
linking to the monorepo/workspace root, instead of wanted package directory.
```

The message assumes a private manifest means the contributor linked the monorepo root by mistake. That holds for monorepos, but it also rejects **single-package repos that set `private: true` at the package root** as a publish guard (they publish a built artifact from CI, not the source root). For these there is no other directory to link, since the root *is* the package, so the entry can never be added or updated.

## Affected packages (single-package repos, root name == published package, no workspaces)

| package | repo | root `name` | root `private` | downloads/mo |
|---|---|---|---|---|
| jotai | pmndrs/jotai | `jotai` | `true` | ~21M |
| zustand | pmndrs/zustand | `zustand` | `true` | ~171M |
| lodash | lodash/lodash | `lodash` | `true` | ~631M |
| @legendapp/state | LegendApp/legend-state | `@legendapp/state` | `true` | ~316K |

Each is published to npm under the same name as its root `package.json`, and none declares `workspaces`. The "you linked the monorepo root" diagnosis does not apply.

## Why it matters

These entries can't receive any field update (platform flags, metadata), because editing an existing entry runs the same check. `zustand` and `@legendapp/state` already carry flags in the directory, so they're stuck at their current data.

## Suggested fix

Skip the `isPrivate` rejection when the resolved npm package name matches the repo's root `package.json` `name`. That confirms the link points at the genuine published package, not a monorepo root. A monorepo root either has no `name` (already caught by the next check) or a `name` that differs from the entry's package, so the existing protection is preserved.

(Alternative: validate against the published manifest from the npm registry rather than the GitHub source manifest.)

Happy to open a PR if this direction sounds right.

mosesroth:
# 📝 Why & how

Fixes #2584.

`validate-new-entries` rejects any entry whose GitHub `package.json` is `private: true`, on the assumption it's a monorepo-root mislink. But several single-package repos (lodash, jotai, zustand, @legendapp/state, …) legitimately keep their root manifest `private: true` as a publish guard while shipping a built artifact via CI, so their Fire OS/other compatibility flags can never be added or updated.

Following the approach from the issue: compare the extracted `package.json` `name` against the `package` field the npm downloads API echoes back (already fetched per entry). When they match, the link points at the genuine published package, so it's allowed past the `isPrivate` guard. A real mislink to an unrelated/monorepo root has a name that differs from (or is absent for) the entry's package and is still rejected by the existing checks.

Changes:
- `fetch-npm-download-data.ts`: include the `package` field in the returned npm data (`validate-new-entries` is its only caller).
- `types/index.ts`: add `npm.package`.
- `validate-new-entries.ts`: relax the `isPrivate` rejection only when extracted name === published name.

Verified locally (`bun ci:validate`):
- lodash (root `private: true`, name `lodash`) now validates.
- a mislink (jotai repo claimed as npmPkg `react`) is still rejected.

# ✅ Checklist

### DIFF
--- a/scripts/fetch-npm-download-data.ts
+++ b/scripts/fetch-npm-download-data.ts
@@ -37,6 +37,7 @@ export async function fetchNpmDownloadData(pkgData: LibraryType, attemptsCount =
       ...pkgData,
       npm: {
         downloads: downloadData.downloads,
+        package: downloadData.package,
       },
     };
   } catch (error) {
--- a/scripts/validate-new-entries.ts
+++ b/scripts/validate-new-entries.ts
@@ -65,11 +65,23 @@ for (let i = 0; i < modifiedEntries.length; i += BATCH_SIZE) {
     }
 
     if (entryWithGitHubData.github.isPrivate === true) {
-      console.error(
-        `Extracted 'package.json' from ${entryWithGitHubData.githubUrl} is marked as private! You might be linking to the monorepo/workspace root, instead of wanted package directory.`
-      );
-      checkResults.push(false);
-      continue;
+      // Some legitimately-published single-package repos keep their root
+      // `package.json` `private: true` as a publish guard (they ship a built
+      // artifact via CI). For those the extracted name still equals the package
+      // npm actually serves, so allow them past this guard. A genuine mislink to
+      // an unrelated/monorepo root has a name that differs from (or is absent
+      // for) the entry's package and is still rejected below.
+      const publishedName = entryWithNpmData.npm?.package;
+      const extractedName = entryWithGitHubData.github.name;
+      const isLegitPrivate = !!extractedName && !!publishedName && extractedName === publishedName;
+
+      if (!isLegitPrivate) {
+        console.error(
+          `Extracted 'package.json' from ${entryWithGitHubData.githubUrl} is marked as private! You might be linking to the monorepo/workspace root, instead of wanted package directory.`
+        );
+        checkResults.push(false);
+        continue;
+      }
     }
 
     if (!entryWithGitHubData.github.name) {


types/index.ts
```diff
@@ -133,6 +133,10 @@ export type LibraryType = LibraryDataEntryType & {
     latestRelease?: string;
     latestReleaseDate?: string;
     hasReadme?: boolean;
+    /** The package name the npm downloads API echoes back (the published name).
+     *  Used by validate-new-entries to compare against the GitHub package.json
+     *  name when allowing legitimately-private single-package repos. */
+    package?: string;
```
Simek: This should not be needed, since we do not use the modified fetcher for main data, and return value of `fetchNpmDownloadData` is not typed as `LibraryType`. Or I'm missing something?

scripts/fetch-npm-download-data.ts
```diff
@@ -37,6 +37,11 @@ export async function fetchNpmDownloadData(pkgData: LibraryType, attemptsCount =
       ...pkgData,
       npm: {
         downloads: downloadData.downloads,
+        // The package name npm actually serves. `validate-new-entries` compares it
+        // against the GitHub `package.json` name to allow legitimately-private
+        // single-package repos (published name === source name) past the
+        // `isPrivate` guard, while still rejecting a mislink to an unrelated repo.
```
Simek:
```suggestion
```

Let's skip the comment, `validate-new-entries` script is the only code which uses that function, and it's quite easy to check what for it's used. I will probably refactor the fetcher a bit, since it has a lot of legacy code in from when it was used for main data fetch.

mosesroth:
Agreed, updated.

Simek (approved):
Thanks for catching this issue and the quick fix! 🙏

PR #2604, opened 2026-07-02
Mark @criipto/verify-expo as supporting the New Architecture

janmeier:
# 📝 Why & how

`@criipto/verify-expo` is an Expo Modules API module. It's architecture-agnostic at runtime and works under the New Architecture. It uses no React Native codegen, so it has no `codegenConfig` for the directory's automatic detection. 

I am the author of the module, and I have manually verified that it works with the new arch :).

Per the README ("Set `newArchitecture` field only when automatic architecture detection fails … despite it supports the New Architecture"), this sets the field manually.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -19524,10 +19524,11 @@
   {
     "githubUrl": "https://github.com/criipto/criipto-verify-expo",
     "npmPkg": "@criipto/verify-expo",
-    "examples": ["https://github.com/criipto/criipto-verify-expo/tree/master/example"],
+    "examples": ["https://github.com/criipto/criipto-verify-expo/tree/master/examples"],
     "ios": true,
     "android": true,
-    "fireos": true
+    "fireos": true,
+    "newArchitecture": true
   },
   {
     "githubUrl": "https://github.com/pagopa/io-app-design-system",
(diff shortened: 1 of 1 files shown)


react-native-libraries.json
```diff
@@ -19524,10 +19524,11 @@
   {
     "githubUrl": "https://github.com/criipto/criipto-verify-expo",
     "npmPkg": "@criipto/verify-expo",
-    "examples": ["https://github.com/criipto/criipto-verify-expo/tree/master/example"],
+    "examples": ["https://github.com/criipto/criipto-verify-expo/tree/master/examples"],
```
Simek: Hey @janmeier, thanks for the update! 👍 

Can you add links per example project? This will enhance the display on the package page.

janmeier: Sure!

janmeier: https://github.com/react-native-community/directory/pull/2604/changes/68a379190a0dc5c8dc662e480a15b7adbd96f08b

It's basically just the same example, with different expo versions. We use them to e2e test our library against different expo versions.

I can also just point it at the expo56 example if you prefer?

Simek: Thanks for the update! 👍 

The idea behind this pattern is that want to have a `package.json` at the root of example so we can run some checks, and extract additional, optional data.

Simek (approved):
LGTM! 🚀

PR #2613, opened 2026-07-05
Fix react-native-nitro-sqlite githubUrl to its monorepo package path

mosesroth:
# 📝 Why & how

The `githubUrl` for **react-native-nitro-sqlite** points to `/tree/main/package`, which doesn't exist — the repo is a monorepo and the published package lives at `packages/react-native-nitro-sqlite`. As a result `validate-new-entries` can't resolve the package's `package.json`.

Corrected the `githubUrl` to the actual package path:

`https://github.com/margelo/react-native-nitro-sqlite/tree/main/packages/react-native-nitro-sqlite`

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -13770,7 +13770,7 @@
     "fireos": true
   },
   {
-    "githubUrl": "https://github.com/margelo/react-native-nitro-sqlite/tree/main/package",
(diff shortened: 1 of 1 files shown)


Simek (approved):
Nice catch, thanks for updating URL after repo refactor! 👍

PR #2671, opened 2026-07-20
Fix Overview crash on hrefless README anchors

### LINKED ISSUE #2670: Package Overview page crashes when a README contains a hrefless `<a>` anchor
## Summary

The package Overview tab crashes and shows the error page ("Uh oh, something went wrong ()") when a package README contains an anchor without an `href`, for example:

```html
<a name="section"></a>
```

## Reproduction

* Open a package whose README contains a hrefless anchor. Example:
  https://reactnative.directory/package/tapflow
* The Overview tab crashes, while the Code / Versions / Score tabs work normally.

Console:

```text
TypeError: Cannot read properties of undefined (reading 'startsWith')
  at MarkdownRenderer.tsx:90
```

## Cause

In `components/Package/MarkdownContentBox/MarkdownRenderer.tsx`:

```ts
if (props.href.startsWith('#')) {
```

`rehype-raw` parses `<a name="...">` as an anchor without an `href`, so `props.href` is `undefined` here and `.startsWith()` throws.

There's already a fallback at the end of the handler:

```tsx
return <span>{props.children}</span>
```

so it looks like this case was supposed to fall through instead of crashing.

## Suggested fix

Using optional chaining here should be enough:

```ts
if (props.href?.startsWith('#')) {
```

Then a hrefless anchor reaches the existing fallback instead of throwing.

Happy to open a PR if this approach looks good.

## Notes

I also reproduced this with a standard `remark -> remark-rehype (allowDangerousHtml) -> rehype-raw` pipeline. A `<a name>` anchor becomes an `<a>` node without an `href`, which matches what happens here.

jo-duchan:
## Summary

Fixes #2670. The package Overview page throws a client-side exception and renders the error page when a README contains a hrefless anchor (e.g. `<a name="section"></a>`).

## Cause

In `MarkdownRenderer.tsx`, the `a` handler starts with:

```ts
if (props.href.startsWith('#')) {
```

`rehype-raw` parses `<a name="...">` into an anchor node with no `href`, so `props.href` is `undefined` and `.startsWith()` throws, which takes down the whole Overview render.

## Fix

```ts
if (props.href?.startsWith('#')) {
```

The handler already ends with a `return <span>{props.children}</span>` fallback, so with optional chaining a hrefless anchor falls through to that fallback instead of crashing. One-character change, no behavior change for anchors that have an `href`.

## Reproduction

Live example before the fix: https://reactnative.directory/package/tapflow (Overview tab crashes; Code / Versions / Score are fine). Console:

```text
TypeError: Cannot read properties of undefined (reading 'startsWith')
  at MarkdownRenderer.tsx:90
```

Also reproduced the parse with a standard `remark -> remark-rehype (allowDangerousHtml) -> rehype-raw` pipeline: a `<a name>` anchor yields one `<a>` node without an `href`, which is what hits the unguarded `.startsWith`.

### DIFF
--- a/components/Package/MarkdownContentBox/MarkdownRenderer.tsx
+++ b/components/Package/MarkdownContentBox/MarkdownRenderer.tsx
@@ -87,7 +87,7 @@ export default function MarkdownRenderer({ data, repoUrl, linkableHeaders = true
         br: () => null,
         hr: () => null,
         a: (props: any) => {
-          if (props.href.startsWith('#')) {
+          if (props.href?.startsWith('#')) {
             return <A {...props} target="_self" />;
           } else if (props.href && !props.href.startsWith('//')) {
             if (!props.href.startsWith('http')) {


Simek (approved):
Thanks for the bug report and fix! 👍

I have fixed it once in the past, but then broke this again when adding support for hash links. 😅 Need to find a way to correctly extract types for custom elements in `@m2d/react-markdown/client` markdown renderer, so we can avoid similar issues in the future.

jo-duchan:
Thanks for the quick merge! Makes sense — those custom-element types are tricky. Glad the optional-chaining fallback helps in the meantime.

PR #2713, opened 2026-08-03
data: mark react-native-livechart as New Architecture compatible

brandtnewlabs:
## Summary

- mark react-native-livechart as new-arch-only in react-native-libraries.json

## Context

react-native-livechart requires react-native-reanimated >=4.0.0. Reanimated 4 supports only the React Native New Architecture (Fabric), so the Directory entry should use the documented new-arch-only value rather than true.

The companion library change commits the monorepo package README so the Directory can discover it:
https://github.com/brandtnewlabs/react-native-livechart/pull/248

## Validation

- JSON parsed successfully with Node
- git diff --check passed

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -22858,6 +22858,7 @@
     "images": [
       "https://raw.githubusercontent.com/brandtnewlabs/react-native-livechart/main/assets/images/hero.webp"
     ],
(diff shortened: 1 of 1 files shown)


Simek (approved):
Hey @brandtnewlabs, thanks for the update! 👍

PR #2717, opened 2026-08-06
Update @kesha-antonov/react-native-action-cable: New Architecture, Expo Go, web, examples

kesha-antonov:
Updating the entry for [`@kesha-antonov/react-native-action-cable`](https://www.npmjs.com/package/@kesha-antonov/react-native-action-cable) (I maintain it).

The library is pure JavaScript with no native code, so automatic New Architecture detection has nothing to detect and the entry currently shows as untested. Adding the flags explicitly, per the note in the README that the field should be set when automatic detection does not return a result.

### Changes

- **`newArchitecture: true`** - the package ships no native code at all. Its only runtime dependency is `eventemitter3`, and the sole reference to `react-native` is a guarded lazy `require` for `AppState` that falls back to a no-op stub when unavailable. Nothing in it touches the bridge, so it runs unmodified on both architectures.
- **`expoGo: true`** - no native module, no config plugin, no `pod install`, so it works in Expo Go without a dev client or prebuild.
- **`web: true`** - works under `react-native-web`; the `AppState` lookup degrades to a no-op when it is not present.
- **`examples`** - two runnable examples in the repo: a full Rails backend plus React Native chat app, and an Apollo GraphQL subscriptions setup.

`fireos` and `vegaos` already resolve to true automatically and are left out, along with all `false` values, per the entry template.


Simek (approved):
Hey @kesha-antonov, thanks for the updates! 👍 

Sidenote: FireOS and Vega OS flag were set by Amazon after testing the library on their side, hence the conflict which I have resolved.

PR #2718, opened 2026-08-06
Update react-native-zoom-reanimated: Expo Go support, explicit npmPkg

kesha-antonov:
Updating the entry for [`react-native-zoom-reanimated`](https://www.npmjs.com/package/react-native-zoom-reanimated) (I maintain it).

### Changes

- **`expoGo: true`** - the library ships no native code. Its `src/` imports only `react`, `react-native`, `react-native-gesture-handler` and `react-native-reanimated`, and both of those native dependencies are bundled in Expo Go, so it runs there without a dev client or prebuild.
- **`npmPkg`** - added explicitly. It matches the repository name so it was being inferred correctly, but stating it keeps the entry stable if that ever stops being true.

`ios`, `android`, `web` and `newArchitecture` are already correct and unchanged.

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -10754,10 +10754,12 @@
   },
   {
     "githubUrl": "https://github.com/kesha-antonov/react-native-zoom-reanimated",
+    "npmPkg": "react-native-zoom-reanimated",
     "examples": ["https://github.com/kesha-antonov/react-native-zoom-reanimated/tree/main/example"],
     "ios": true,
     "android": true,
     "web": true,
+    "expoGo": true,
     "newArchitecture": true
   },
   {


Simek (approved):
Hey @kesha-antonov, thanks for the updates! 👌 

When it comes to `npmPkg`, the next libraries cleanup will wipe it out automatically from the definition file, but can stay in for now, since it does not affect anything. If missing, the field is autogenerated anyway, and when specified it's also validated if it's correct with `package.json` content.

Simek:
This makes me think if we need that field anyway, it's a nice guard to catch incorrect entries, but maybe the extraction can figure it out automatically. Will think about it.

PR #2746, opened 2026-08-19
Add three New Architecture native modules

AndrewDongminYoo:
# 📝 Why & how

Adds three React Native native modules I maintain. All are published to npm, MIT,
ship `codegenConfig`, and have an example app in the repository.

| Package | What it wraps |
| --- | --- |
| `react-native-receipt-scanner` | ML Kit document scanner, for capturing a receipt as a cropped, deskewed image |
| `react-native-step-counter-newarch` | Android `StepCounter`/`StepDetector` sensors and iOS `CMPedometer` |
| `react-native-naver-login-turbo` | Naver OAuth 2.0 login |

Each is marked `"newArchitecture": "new-arch-only"`: they are TurboModule/Fabric
implementations with no legacy Bridge path, so a host app needs `newArchEnabled=true`.
All three contain custom native code and ship no Expo config plugin, so `expoGo` is
not set.

Two entries carry `npmPkg` because the published package name differs from the
repository name.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -23918,5 +23918,28 @@
     "ios": true,
     "android": true,
     "dev": true
+  },
+  {
+    "githubUrl": "https://github.com/AndrewDongminYoo/react-native-receipt-scanner",
+    "examples": ["https://github.com/AndrewDongminYoo/react-native-receipt-scanner/tree/main/example"],
+    "ios": true,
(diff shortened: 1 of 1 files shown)


Simek (approved):
Hey @AndrewDongminYoo, thanks for adding new entries to the directory. 👍

PR #2754, opened 2026-08-23
Update Nitro Image package metadata

mrousavy:
# 📝 Why & how

Updates both Nitro Image packages with their shared example application and correct New Architecture status:

- `react-native-nitro-image` uses a Nitro View, which requires the New Architecture.
- `react-native-nitro-web-image` depends on `react-native-nitro-image`, so the package stack is also New Architecture only.

This replaces the currently inferred untested status with the documented compatibility status for both packages.

Validated with:

- `bun data:test`
- `bun data:validate`
- `bun ci:validate`
- `bun lint`

# ✅ Checklist

- [x] Updated libraries in **`react-native-libraries.json`**

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -19029,9 +19029,11 @@
   },
   {
     "githubUrl": "https://github.com/mrousavy/react-native-nitro-image/tree/main/packages/react-native-nitro-image",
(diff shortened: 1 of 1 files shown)


Simek (approved):
Thanks! 👌

PR #2756, opened 2026-08-23
Update VisionCamera v5 metadata

mrousavy:
# 📝 Why & how

Updates the VisionCamera domain for the v5 Nitro rewrite:

- Marks `react-native-vision-camera` and all five v5 companion packages as New Architecture only.
- Removes the obsolete legacy-view compatibility note from the core package.
- Removes `configPlugin` because VisionCamera v5 no longer ships an Expo config plugin.
- Adds the shared `apps/simple-camera` example to each companion package. The app installs and exercises the barcode scanner, location, resizer, Skia, and worklets packages.

VisionCamera v5 uses Nitro Views, which require the New Architecture. Each companion package depends on the v5 core package.

Validated with:

- `bun data:test`
- `bun data:validate`
- `bun ci:validate`
- `bun lint`

# ✅ Checklist

- [x] Updated libraries in **`react-native-libraries.json`**

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -7105,9 +7105,7 @@
     "ios": true,
     "android": true,
     "fireos": true,
-    "newArchitecture": true,
-    "newArchitectureNote": "Currently supported through the compatibility layer for legacy view components.",
-    "configPlugin": true
+    "newArchitecture": "new-arch-only"
   },
   {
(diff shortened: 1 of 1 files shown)


Simek (approved):
Thanks for the updates @mrousavy! 🚀

PR #2765, opened 2026-08-27
Add a demo image for expo-tts-file

ksblazh:
Adds an `images` entry for [`expo-tts-file`](https://github.com/ksblazh/expo-tts-file), merged into the directory a couple of days ago in #2764.

The recording shows what the package does on a device: it synthesizes text to an audio file, then plays that file back with each word highlighted in time with the speech. The highlighting is driven by word timings returned alongside the file rather than by live progress events, which is why it still lines up on replay and on seek.

Hosted from the package's own repository, so nothing new is added to this repo.


Simek (approved):
Thanks for the update! 🚀

PR #2769, opened 2026-08-29
Add munim-ffmpeg

sheehanmunim:
Adds [munim-ffmpeg](https://github.com/munimtechnologies/munim-ffmpeg), FFmpeg and FFprobe for React Native and Expo, built on Nitro Modules.

- Ships **FFmpeg 9.0.1** built from ffmpeg.org for both platforms — iOS arm64 plus arm64/x86_64 simulator, Android arm64-v8a, armeabi-v7a and x86_64 (16 KB aligned).
- New architecture only, via Nitro Modules; includes an Expo config plugin.
- LGPL: no x264/x265, so apps do not inherit GPL obligations. H.264/HEVC use VideoToolbox and MediaCodec, with openh264 as a software fallback.
- Example app runs a 25-check device suite; verified on an iPad Air (M3), the iOS Simulator and a Galaxy A14 5G.

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -22548,6 +22548,17 @@
     "configPlugin": true
   },
   {
+    "githubUrl": "https://github.com/munimtechnologies/munim-ffmpeg",
+    "npmPkg": "munim-ffmpeg",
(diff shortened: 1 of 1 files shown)


Simek (approved):
Hey @sheehanmunim, thanks for adding a new package to the directory! 👍

PR #2763, opened 2026-08-26
Add rn-story

AbdullahAnsarii:
# 📝 Why & how

Adds [rn-story](https://github.com/AbdullahAnsarii/rn-story) ([npm](https://www.npmjs.com/package/rn-story)) — a lightweight, TypeScript-first component for Instagram/WhatsApp/Snapchat-style stories. Image and video stories with animated progress bars, tap-to-navigate and long-press-to-pause gestures, per-story "See More" links, and a custom header slot. Works with Expo (expo-av peer dependency); v2.0.1 also runs on react-native-web.

Entry appended at the end of `react-native-libraries.json` per the contribution guide, with a screenshot and a link to the bundled example app.

# ✅ Checklist

- [x] Added library to **`react-native-libraries.json`**

PR #2779, opened 2026-09-01
Mark rn-story as supporting the New Architecture

AbdullahAnsarii:
# 📝 Why & how

Sets `newArchitecture: true` on the existing `rn-story` entry (added in #2763).

rn-story is pure JavaScript with no native code of its own, so it has nothing architecture-specific to migrate. Its only native dependency — `expo-av`, consumed as a peer dependency — runs under the New Architecture on the Expo SDKs that ship it (SDK 52+, where the New Architecture is the default). The library's example app and Snack demo target SDK 54.

This clears the "?" New Architecture badge on the listing and the `expo doctor` "Untested on New Architecture" warning for projects using the package.

# ✅ Checklist

### DIFF
--- a/react-native-libraries.json
+++ b/react-native-libraries.json
@@ -24087,7 +24087,8 @@
     "images": ["https://raw.githubusercontent.com/AbdullahAnsarii/rn-story/master/docs/demo.jpg"],
     "ios": true,
     "android": true,
-    "web": true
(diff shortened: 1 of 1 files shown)


Simek (approved):
Thanks for the update! 👍

PR #2785, opened 2026-09-03
Fix empty query on route push

cornejobarraza:
# 📝 Why & how

### Why

Typing in the search box and pressing Enter right after causes the route push to have an empty query

### How

Added a search ref to be used exclusively by the route push to not impact debounced API call search

### Result

https://github.com/user-attachments/assets/5afd8fd5-650c-489b-add2-206bbee91c81

# ✅ Checklist


Simek (approved):
Nice catch, and clean fix, thanks @cornejobarraza! 👍

PR #2788, opened 2026-09-07
Update @lugg/maps to new-arch-only

lodev09:
# 📝 Why & how

`@lugg/maps` is a Fabric-only library (see [installation docs](https://maps.lodev09.com/installation)). Set `newArchitecture` to `"new-arch-only"` instead of `true`.

# ✅ Checklist


Simek (approved):
Hey @lodev09, thanks for the update! 👍

